PR #14, opened 2019-09-02
Extend Windows Support

DavidMansolino:
Improve support of Windows.

Doc updated too: http://wiki.ros.org/webots_ros2#From_Sources

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,9 @@ The following procedure should be used to compile the package and launch the sim
 
 ```
 export WEBOTS_HOME=~/Downloads/webots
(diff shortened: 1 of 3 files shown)


DavidMansolino:
I have tried it on Windows using this installation procedure: https://index.ros.org/doc/ros2/Installation/Dashing/Windows-Install-Binary/

I managed to make most of our packages compile and run on Windows except from the ``webots_ros2_msgs`` one, but the problem is not on our side, the generation of services and messages doesn't seem to work in the latest Windows release of ROS2.

PR #19, opened 2019-09-06
Feature Demo

DavidMansolino:
The goal of this PR is to create a nice demo where an universal robot is simulated in Webots, controlled with ROS2 and displayed in RVIZ.

## For testing only

Testing the tf publisher: `ros2 run webots_ros2_core tf_publisher --webots-robot-name=tf_supervisor`

Publishing the robot model: `ros2 topic pub robot_description std_msgs/String "{ data: /home/cyberbotics/ros2_workspace/install/webots_ros2_ur_e_description/share/webots_ros2_ur_e_description/urdf/ur5e_robot.urdf }"`

With launch instead: `ros2 launch webots_ros2_universal_robot universal_robot_rviz.launch.py`


DavidMansolino:
Current state: I can send the urdf definition in RVIZ, share position/orientation of each part between Webtos and RVIZ, but for some reason RVIZ is not able to load the meshes yet.

DavidMansolino:
Screenshot from 2019-09-06 12-07-04

:smiley:

DavidMansolino:
Remaining issues:
  - [ ] If webots/lib is in the path RVIZ doesn't work, and if it is not the controllers don't work (I am really having a hard time with this one).
  - [ ] Remove hard-coded path of the 'urdf' file in the RVIZ config file (this should be possible using a topic, but doesn't work fine in ROS2 yet).

DavidMansolino:
First issue is nicely fixed with cyberbotics/webots#879.
The second problem seems to be on the RVIZ2 side (https://answers.ros.org/question/304645/ros2-robot-state-publisher-publishes-empty-robot_description/) I tried many option but none of them work.

DavidMansolino:
:+1: 

(RVIZ will launch only if tested from cyberbotics/webots#879)

PR #22, opened 2019-09-06
Fix webots_ros2_desktop package creation

### LINKED ISSUE #21: webots_ros2_desktop miss Webots
When compiled from the build farm of ROS2 the is missing Webots (compilation succeeded but generated message does not include Webots).

We should extend this script: https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_desktop/setup.py#L9

To download Webots and include it.

DavidMansolino:
On the ROS2 build farm Webots is not available, so we need to download and extract it during the compilation process.

Fix #21.


DavidMansolino:
:+1:

webots_ros2_desktop/setup.py
```diff
@@ -1,12 +1,35 @@
...
+import urllib.request
 
 from setuptools import setup
 
 package_name = 'webots_ros2_desktop'
 data_files = []
+# If 'WEBOTS_HOME' not set try to download latest package (only on linux)
+if 'WEBOTS_HOME' not in os.environ and 'TRAVIS' not in os.environ and sys.platform == 'linux':
+    # Remove previous archive
+    archiveName = 'webots-R2019b-x86-64.tar.bz2'
```
omichel: It may be better to put the Webots version in a config file instead of hardcoding it here and there.

DavidMansolino:
:+1:

PR #39, opened 2019-10-15
Fix Node Exit

DavidMansolino:
When the world is reverted the node is not exiting correctly, it somehow remains running (even if the main loop is not running anymore).


lukicdarkoo:
It should not happen with `webots_ros2_driver` after #301 is merged

PR #46, opened 2019-12-19
Update to Webots R2020a

DavidMansolino:
Update to the latest version of Webots.


webots_ros2_desktop/setup.py
```diff
@@ -33,6 +34,13 @@
     tar.close()
     os.environ['WEBOTS_HOME'] = os.path.abspath(os.path.join(os.path.dirname(__file__),
                                                              'webots'))
+    # remove dead symlink in R2020a
+    os.remove(os.path.join(os.environ['WEBOTS_HOME'], 'lib', 'controller', 'libssl.so'))
+    os.remove(os.path.join(os.environ['WEBOTS_HOME'], 'lib', 'controller', 'libcrypto.so'))
+    # copy required library
+    copyfile(os.path.join(os.environ['WEBOTS_HOME'], 'lib', 'webots', 'libpng12.so.0'),
+             os.path.join(os.environ['WEBOTS_HOME'], 'lib', 'controller', 'libpng12.so.0'))
```
DavidMansolino: For some strange reason of the ros nodes fails to import Supervisor if this library is not copied in the lib/controller folder (but using the Supervisor API in an extern controller outside of ROS2 works fine).

DavidMansolino: Ok, this seems normal as libcontroller has a dependency on libpng12.so (which is not available on Ubuntu 18.04). So let's keep it like this as long as it is not fixed in Webots.

DavidMansolino:
:+1:

PR #63, opened 2020-04-07
Integrate Navigation2 to e-puck example

lukicdarkoo:
**Description**
We would like to integrate ROS2 Navigation2 stack for e-puck so e-puck can autonomously navigate to the given pose. This will allow us to demonstrate the benefits of ROS2 integration and to build multiple demos upon the Navigation2.

**Affected Packages**
  - webots_ros2_epuck

**Tasks**
  - [x] Create a node that publishes an empty map
  - [x] Add Navigatio2 and the node (that publishes the map) to launch file
  - [x] Fix `[ERROR] [controller_server]: No valid trajectories out of 0!` (fixed basing the configuration default Controller Server configuration instead of the recommended one)
  - [x] Improve Navigation2 Controller server configuration
    - [x] Precision, the robot should move closer to the goal
    - [x] Obstacle avoidance should be better
    - [x] Sometimes the robot stops, it gets lost [1]
  - [x] Build the main map based on sensor readings
  - [x] Separate the launch file so it can be easily used for the physical robot
  - [x] Not directly related bug fixes
    - [x] Gyroscope sends data only if it exists as a node
    - [x] Gyroscope data is scaled properly so it matches the raw values
    - [x] Velocity in odometry is more precise due to timing changes

**Small Tasks**
  - [x] Add necessary dependencies to package.xml
  - [x] Add documentation

**Additional Context**
- RViz may not show a cost map depending on the version (with `ff8fcf9a2411` works just fine): `https://github.com/ros-planning/navigation2/issues/921`
- [1] Progress checker kicks in after 10s and "says" to the controller that the robot didn't manage to pass 0.5m in 10s. In general, it is fine, but since our robot is small and slow it always raises the error after 10s and enters in the recovery mode. Unfortunately, there is a bug in Eloquent which forbids these parameters to be changed: https://answers.ros.org/question/344004/configuring-the-progress-checker-in-navigation2/

**Result**
image
[Check the video](https://i.imgur.com/A4kloOj.mp4)


webots_ros2_epuck/webots_ros2_epuck/driver.py
```diff
@@ -281,6 +281,7 @@ def step_callback(self):
         self.publish_distance_data(stamp)
```
DavidMansolino, line 281: This Is not directly related to this PR, but I just saw a small issue at line 101, this is duplicated:
```
self.camera_period = camera_period_param.value
```

DavidMansolino:
👀

webots_ros2_core/webots_ros2_core/webots_differential_drive_node.py
```diff
@@ -58,43 +56,40 @@ def __init__(self,
...
         self.left_wheel_sensor.enable(self.timestep)
         self.right_wheel_sensor.enable(self.timestep)
-        self._odometry_publisher = self.create_publisher(
-            Odometry, odometry_topic, 1)
+        self._odometry_publisher = self.create_publisher(Odometry, odometry_topic, 1)
         self._tf_broadcaster = TransformBroadcaster(self)
 
         # Initialize timer
-        self.create_timer(self.timestep / 1000.0, self._publish_odometry_data)
+        self._last_sample_time = self.robot.getTime()
```
DavidMansolino, line 71: I would add the `odometer` keyword in this variable so that we know what this sample time refer to.

webots_ros2_epuck/webots_ros2_epuck/simple_mapper.py
```diff
@@ -0,0 +1,161 @@
...
+        # Init map related elements
+        self.map = [-1]*MAP_WIDTH*MAP_HEIGHT
+        self.map_publisher = self.create_publisher(
+            OccupancyGrid,
+            '/map',
+            qos_profile=QoSProfile(
+                depth=1,
+                durability=DurabilityPolicy.TRANSIENT_LOCAL,
+                history=HistoryPolicy.KEEP_LAST,
+            )
```
DavidMansolino, line 53: Just for the curiosity, why the default QoS profile was not working?

webots_ros2_epuck/README.md
```diff
@@ -129,16 +129,74 @@ Ground sensors come as an [optional module](http://www.e-puck.org/index.php?opti
...
+    z: 0.0
+    w: 1.0
+"
+```
+
+![Demo](./assets/nav2.gif)
+
+This example will work properly only for [ROS2 Foxy](https://index.ros.org/doc/ros2/Releases/Release-Foxy-Fitzroy/) (the first ROS2 release that has a long support - 3+ years):
+- Navigation2 stack is tested with version `e3469486675beb3` that includes [fix of progress checker parameters](https://answers.ros.org/question/344004/configuring-the-progress-checker-in-navigation2/) and [namespaced plugins for servers](https://github.com/ros-planning/navigation2/pull/1468).
+- RViz2 for Eloquent has a bug and [cannot show a local cost map](https://github.com/ros-planning/navigation2/issues/921), therefore, `ff8fcf9a2411` (or up) version of RViz2 is desired.
```
DavidMansolino, line 181: Is there any possibility to test this PR without having to recompile any package manually?

webots_ros2_epuck/webots_ros2_epuck/simple_mapper.py
lukicdarkoo: I wanted to make it "latched" topic, so "late" subscribers get the last map from the buffer - they don't have to wait for the new map to be published.

webots_ros2_epuck/README.md
lukicdarkoo: > I am wondering about the `example_tools_launch` and `simple_mapper, they have nothing specific to the e-puck right?
> In that case, maybe we can think about putting them in a more generic package, what do you think?

Hmm, maybe, RViz2 and Navigation2 configuration files (Navigation2 configuration file is really tuned for e-puck) are made for e-puck. Therefore, those two files should be parameters. `simple_mapper` doesn't make sense for robots that have proper sensors, but that is already optional.

In short, with RViz2 and Navigation2 configurable (config filepath) it would work.

lukicdarkoo:
> I am wondering about the `example_tools_launch` and `simple_mapper, they have nothing specific to the e-puck right?
> In that case, maybe we can think about putting them in a more generic package, what do you think?

Hmm, maybe, RViz2 and Navigation2 configuration files (Navigation2 configuration file is really tuned for e-puck) are made for e-puck. Therefore, those two files should be parameters. `simple_mapper` doesn't make sense for robots that have proper sensors, but that is already optional.

In short, with RViz2 and Navigation2 configurable (config file path) it would work.

DavidMansolino:
> > I am wondering about the `example_tools_launch` and `simple_mapper, they have nothing specific to the e-puck right?
> > In that case, maybe we can think about putting them in a more generic package, what do you think?
> 
> Hmm, maybe, RViz2 and Navigation2 configuration files (Navigation2 configuration file is really tuned for e-puck) are made for e-puck. Therefore, those two files should be parameters. `simple_mapper` doesn't make sense for robots that have proper sensors, but that is already optional.
> 
> In short, with RViz2 and Navigation2 configurable (config file path) it would work.

Ok, I forgot about the configuration files. In that case the interest to move it in a more global package is way smaller. In that case let's keep it here for now and if a some point we need it in another package (e.g. the Khepera IV one) we can move it (as long as it is not officially released we can afford moving it without any problem).

webots_ros2_epuck/README.md
DavidMansolino: Ok in that case I will recompile them.

DavidMansolino:
👀

DavidMansolino:
I was finally able to recompile and run everything on my VM.
It is working really nicely, that's a very nice demo!

Just a few comments:
It seems it most of the times have issues finding obstacles that are seen by the short distance sensors:
Screenshot 2020-04-16 at 13 57 34
Do you know why?

Have you tried running this navigation stack with the real robot too?

And some comments not linked with this PR (you might fix this in another PR if you prefer):
  - The e-puck world contains a R2020b header, which is causing a warning when opening with the current version of Webots (R2020a): https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_epuck/worlds/epuck_world.wbt#L1
  - It would be nice to set the resolution of the camera of the simulated robot bigger (even maybe the same as the real one) in the world file.
  - In this sample I can see that the camera is enabled even if nothing subscribes to the topic, would it be possible to disable the camera if nobody subscribed (not needed for the other sensors way less computationally expensive)?

lukicdarkoo:
> Do you know why?

I reduced the range of infrared sensors to 0.009-0.04m (to better match the sensors on the real robot), so maybe that, or low update rate of navigation stack, I will check and try to fix.

> In this sample I can see that the camera is enabled even if nothing subscribes to the topic, would it be possible to disable the camera if nobody subscribed (not needed for the other sensors way less computationally expensive)?

Sure! I can do the same for all sensor, do you think if it is a good idea?

webots_ros2_epuck/README.md
lukicdarkoo: BTW, I am sketching ROS2 package manager that should be able to install packages from sources more easily:
https://gist.github.com/lukicdarkoo/d03196b15367b804b27c142dad8644e9

It may be helpful in similar situations

DavidMansolino:
> same for all sensor, do you think if it is a good idea?

If it doesn't complexity too much the code yes, otherwise camera is the most important one (because it is the only one really computationally expensive).

> Yes, it works with a real robot, although there is a space for improvement for both, simulation and real robot navigation. It's part of my project to benchmark performances of the real and simulated robot, so I thought to make it better then. E.g., sometimes it happens that robot cannot find a path even though it is possible for him to pass between obstacles.

Very good!

DavidMansolino:
👍

webots_ros2_epuck/worlds/epuck_world.wbt
```diff
@@ -1,9 +1,9 @@
-#VRML_SIM R2020a utf8
+#VRML_SIM R2020b utf8
```
DavidMansolino: Be careful, you re-introduced the wrong header in this commit 😉

lukicdarkoo: Oh, no :facepalm:

lukicdarkoo:
> The enabling/disabling of the camera and other sensors is very cool.
I didn't include distance sensors only because it would create messy code and those sensors are often used anyways.

lukicdarkoo:
> It seems it most of the times have issues finding obstacles that are seen by the short distance sensors

I have just tested the navigation with commit [`5fd07b47e7770206276f`](https://github.com/ros-planning/navigation2/commit/5fd07b47e7770206276f9d49e0921d4bd819fc15) and it discovers the obstacles much faster, and in general, the navigation works now.

PR #69, opened 2020-04-22
Separate Common and Webots Specific Documentation for e-puck

lukicdarkoo:
**Description**
Packages `epuck_ros2_cpp` is missing documentation. Since most of the documentation is shared this PR will organise the docs files properly.

**Related Issues**
https://github.com/cyberbotics/epuck_ros2/issues/9

**Affected Packages**
  - webots_ros2_epuck

**Tasks**
  - [x] Separate common and Webots specific documentation
  - [x] Rename bring up launch file to `robot_launch.py` according to the convention from turtlebot (https://github.com/ROBOTIS-GIT/turtlebot3/tree/ros2/turtlebot3_bringup/launch)


webots_ros2_epuck/EPUCK_ROS2.md
```diff
@@ -0,0 +1,206 @@
+# ROS2 for E-Puck
+Here you will find instructions on how to use utilize e-puck ROS2 API.
```
DavidMansolino, line 2:
```suggestion
Here you will find instructions on how to use the e-puck ROS2 API. 
```

webots_ros2_epuck/EPUCK_ROS2.md
```diff
@@ -0,0 +1,206 @@
+# ROS2 for E-Puck
+Here you will find instructions on how to use utilize e-puck ROS2 API. 
+
+This documentation is common for for physical ([`epuck_ros2`](https://github.com/cyberbotics/epuck_ros2)) and simulated ([`webots_ros2_epuck`](https://github.com/cyberbotics/webots_ros2/tree/master/webots_ros2_epuck)) robot, so there is a slight difference in using the launch file.
```
DavidMansolino, line 4:
```suggestion
This documentation is common for the physical ([`epuck_ros2`](https://github.com/cyberbotics/epuck_ros2)) and simulated ([`webots_ros2_epuck`](https://github.com/cyberbotics/webots_ros2/tree/master/webots_ros2_epuck)) robot, so there is a slight difference in using the launch file.
```

webots_ros2_epuck/EPUCK_ROS2.md
```diff
@@ -0,0 +1,206 @@
...
+Launch file `example_launch.py` runs the Webots simulation and includes `example_tools_launch.py`.
+Therefore, you can use the launch file `example_tools_launch.py` for the physical robot with the same parameters as you would `example_launch.py` for the simulated robot.
+To emphasis this, you will see `[tools_]` in the commands which indicates you can use `example_tools_launch.py` and `example_launch.py` depending on whether you work with the physical or simulated robot, e.g.:
```
DavidMansolino, line 7: I am not sure to understand why the name of the launch files are different, since the package names are not the same there should be no name clash if the launch files use the same names no?

webots_ros2_epuck/README.md
```diff
@@ -1,252 +1,10 @@
-## Run demo
+# ROS2 Driver for E-Puck Robot in Webots
+This package adds ROS2 support for [e-puck](https://www.gctronic.com/doc/index.php/e-puck2) simulated robot in [Webots](https://cyberbotics.com/). Since ROS2 driver for physical ([`epuck_ros2`](https://github.com/cyberbotics/epuck_ros2)) and simulated ([`webots_ros2_epuck`](https://github.com/cyberbotics/webots_ros2/tree/master/webots_ros2_epuck)) robot share the same API you can find more instructions [here](./EPUCK_ROS2.md).
```
DavidMansolino, line 2:
```suggestion
This package adds ROS2 support for [e-puck](https://www.gctronic.com/doc/index.php/e-puck2) simulated robot in [Webots](https://cyberbotics.com/).
Since ROS2 driver for the physical ([`epuck_ros2`](https://github.com/cyberbotics/epuck_ros2)) and simulated ([`webots_ros2_epuck`](https://github.com/cyberbotics/webots_ros2/tree/master/webots_ros2_epuck)) robot share the same API you can find more instructions [here](./EPUCK_ROS2.md). 
```

webots_ros2_epuck/EPUCK_ROS2.md
```diff
@@ -0,0 +1,206 @@
...
+```
+ros2 launch webots_ros2_epuck example_[tools_]launch.py rviz:=true
+```
+
+## Infra-red, Light Sensors, and LEDs
+![e-puck2 camera and infrared sensors](https://raw.githubusercontent.com/cyberbotics/webots/master/docs/guide/images/robots/epuck/sensors_and_leds.png)
```
DavidMansolino, line 13: I would rather re-upload this image in this repo as it might easily be broken by changes in the other repo.

webots_ros2_epuck/EPUCK_ROS2.md
lukicdarkoo: First, you have to run driver:
- `ros2 launch webots_ros2_epuck example_launch.py` for simulated robot or
- `ros2 launch epuck_ros2 example_launch.py` for physical robot.

Second, you can run extra tools/demos with:
`ros2 launch webots_ros2_epuck example_tools_launch.py`
(this works for both, simulated and physical robot)

For convenience, `webots_ros2_epuck/example_launch.py` also includes `webots_ros2_epuck/example_tools_launch.py` so you don't need to run two launch files.

But since you didn't understand it is either poorly explained or organized. Do you have any idea how to simplify?

DavidMansolino: > First, you have to run driver:
> 
> * `ros2 launch webots_ros2_epuck example_launch.py` for simulated robot or
> * `ros2 launch epuck_ros2 example_launch.py` for physical robot.

Ok perfect, this is indeed what I would expect that behaviour and name of launch files are exactly the same for real and simulated robot (juste the package name changes).

> Second, you can run extra tools/demos with:
> `ros2 launch webots_ros2_epuck example_tools_launch.py`
> (this works for both, simulated and physical robot)
> 
> For convenience, `webots_ros2_epuck/example_launch.py` also includes `webots_ros2_epuck/example_tools_launch.py` so you don't need to run two launch files.
> 
> But since you didn't understand it is either poorly explained or organized. Do you have any idea how to simplify?

Ok, after reading it several times I understand know the difference, the difference is that example_launch.py doesn't not start the tools on the real robot but on the simulated one yes, am I right?

I think this is the perturbing point, I am wondering if we should not make the `example_launch.py` work and behave the exact same way with real robot and in simulation. And add an extra launch (e.g. `example_with_tools_launch.py` that will do what the current ``example_launch.py`` is doing in simulation). What do you think?

lukicdarkoo: I guess it is a good idea, we can proceed that!

lukicdarkoo: I think it is clear now. I also renamed `example_launch.py` to `robot_launch.py` to be closer to turtlebot convention

DavidMansolino: Much clearer indeed, thank you!

PR #74, opened 2020-04-23
Replace .now() method

### LINKED ISSUE #71: Improper conversion to rclpy.Time for nanoseconds
In `webots_node`, time is converted from unix timestamp format to `rclpy.Time`

https://github.com/cyberbotics/webots_ros2/blob/40385d93f49812357145cb9b285758b8fcbcbf55/webots_ros2_core/webots_ros2_core/webots_node.py#L93

However, there seems to be an error in the conversion as stated nanoseconds are in fact microseconds. Therefore, the line should be :

```python
nanoseconds = int((sim_time - seconds) * 1e9)
```

lukicdarkoo:
**Description**
Nanoseconds are not calculated properly.

**Related Issues**
Fixes #71

**Affected Packages**
  - webots_ros2_core
  - webots_ros2_epuck
  - webots_ros2_demos

**Tasks**
  - [x] Decide wether to deprecate `.now()` method in favor of `self.get_clock().now().to_msg()` or to fix it
  - [x] Fix time usage in:
    - [x] `driver.py` (epuck), 
    - [x] `trajectory_follower.py`,
    - [x] `follow_joint_trajectory_client.py`,
    - [x] `tf_publisher.py` and
    - [x] `laser_publisher.py` (not tested)
  - [x] Increase line length in `webots_ros2_core`

**Additional Context**
- Users can use either `Time(seconds=self.robot.getTime()).to_msg()` or `self.get_clock().now().to_msg()`

### DIFF
--- a/webots_ros2_core/webots_ros2_core/joint_state_publisher.py
+++ b/webots_ros2_core/webots_ros2_core/joint_state_publisher.py
@@ -19,7 +19,7 @@ import sys
 from webots_ros2_core.utils import append_webots_python_lib_to_path
 
 from sensor_msgs.msg import JointState
-from builtin_interfaces.msg import Time
+from rclpy.time import Time
 
 try:
     append_webots_python_lib_to_path()
@@ -57,10 +57,8 @@ class JointStatePublisher():
 
     def publish(self):
         """Publish the 'joint_states' topic with up to date value."""
-        seconds = int(self.robot.getTime())
-        nanoseconds = int((self.robot.getTime() - seconds) * 1.0e+6)
         msg = JointState()
-        msg.header.stamp = Time(sec=seconds, nanosec=nanoseconds)
+        msg.header.stamp = Time(seconds=self.robot.getTime()).to_msg()
         msg.header.frame_id = 'From simulation state data'
         msg.name = [s + self.jointPrefix for s in self.jointNames]
         msg.position = []
--- a/webots_ros2_core/webots_ros2_core/trajectory_follower.py
+++ b/webots_ros2_core/webots_ros2_core/trajectory_follower.py
@@ -22,7 +22,8 @@ from webots_ros2_core.utils import append_webots_python_lib_to_path
 
 from control_msgs.action import FollowJointTrajectory
 from trajectory_msgs.msg import JointTrajectoryPoint
-from builtin_interfaces.msg import Duration
+from rclpy.duration import Duration
+from rclpy.time import Time
 
 from rclpy.action import ActionServer, CancelResponse, GoalResponse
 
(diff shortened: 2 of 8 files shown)


DavidMansolino:
Note directly related to ``now()`` but the wrong 1.0e+6 factor is also used here:
  - https://github.com/cyberbotics/webots_ros2/blob/28db7084deee68bc26d64b89a687c7c28584147b/webots_ros2_core/webots_ros2_core/trajectory_follower.py#L85
  - https://github.com/cyberbotics/webots_ros2/blob/28db7084deee68bc26d64b89a687c7c28584147b/webots_ros2_core/webots_ros2_core/trajectory_follower.py#L104
  - https://github.com/cyberbotics/webots_ros2/blob/28db7084deee68bc26d64b89a687c7c28584147b/webots_ros2_demos/webots_ros2_demos/follow_joint_trajectory_client.py#L73
  - https://github.com/cyberbotics/webots_ros2/blob/0706b89be6f96491b7a152b6c5fa35de65705de4/webots_ros2_core/webots_ros2_core/tf_publisher.py#L49
  - https://github.com/cyberbotics/webots_ros2/blob/0706b89be6f96491b7a152b6c5fa35de65705de4/webots_ros2_core/webots_ros2_core/laser_publisher.py#L95
  - https://github.com/cyberbotics/webots_ros2/blob/28db7084deee68bc26d64b89a687c7c28584147b/webots_ros2_core/webots_ros2_core/trajectory_follower.py#L104

Some even looks wrong (not only the factor).

webots_ros2_core/webots_ros2_core/joint_state_publisher.py
```diff
@@ -57,10 +57,8 @@ def __init__(self, robot, jointPrefix, node):
 
     def publish(self):
         """Publish the 'joint_states' topic with up to date value."""
-        seconds = int(self.robot.getTime())
-        nanoseconds = int((self.robot.getTime() - seconds) * 1.0e+6)
         msg = JointState()
-        msg.header.stamp = Time(sec=seconds, nanosec=nanoseconds)
+        msg.header.stamp = Time(seconds=self.robot.getTime()).to_msg()
```
DavidMansolino, line 61: It's probably ok, but are you sure that ``seconds`` handles floats?

webots_ros2_core/webots_ros2_core/trajectory_follower.py
```diff
@@ -114,16 +109,14 @@ def sample_trajectory(trajectory, t):
...
     # Last point
-    if t >= (trajectory.points[-1].time_from_start.sec +
-             trajectory.points[-1].time_from_start.nanosec * 1.0e-6):
+    if t >= Duration.from_msg(trajectory.points[-1].time_from_start):
         return copy.deepcopy(trajectory.points[-1])
     # Finds the (middle) segment containing t
     i = 0
-    while (trajectory.points[i + 1].time_from_start.sec +
-           trajectory.points[i + 1].time_from_start.nanosec * 1.0e-6) < t:
+    while Duration.from_msg(trajectory.points[i + 1].time_from_start) < t:
```
DavidMansolino, line 119: Much cleaner this way! Thanks

webots_ros2_core/webots_ros2_core/joint_state_publisher.py
lukicdarkoo:
```
$ python3 -c 'from rclpy.time import Time; print(Time(seconds=0.32))'
Time(nanoseconds=320000000, clock_type=SYSTEM_TIME)
```
Yes, `Time` and `Duration` work nicely. Although I am disappointed there is no attribute `seconds` as I would like to do:
```
Time(nanoseconds=123).seconds
```
They may that, I checked their implementation and it should be straightforward.

DavidMansolino: > They may that, I checked their implementation and it should be straightforward.

If it is really annoying, we might create classes that inherits from these classes and extend them, but this is probably a bit overkilled.

PR #77, opened 2020-04-27
Add Generic Differential Drive Node

### LINKED ISSUE #73: Generic Differential Drive Node
Now that we do provide a generic class (https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_core/webots_ros2_core/webots_differential_drive_node.py) to create differential drive nodes, it would be nice to slightly extend it to include also a generic differential drive node.

For this we would need to add a main to this class, something like (untested draft):
```Python

def main(args=None):
    rclpy.init(args=args)

    # TODO: we need to get 'wheel_distance' and 'wheel_radius' from the ros params.
    differential_drive = WebotsDifferentialDriveNode(args=args, wheel_distance, wheel_radius)

    rclpy.spin(differential_driver)
    rclpy.shutdown()


if __name__ == '__main__':
    main()
```

This will provide a very simple (and limited) ros node allowing to control any differential drive out of the box.

lukicdarkoo:
**Description**
This has to allow the differential drive node to be used as a standalone node.

**Related Issues**
Closes #73

**Affected Packages**
  - webots_ros2_core
  - webots_ros2_tiago

**Tasks**
  - [x] Allow `WebotsDifferentialDriveNode` to initialized in launch file 
  - [x] Integrate it into Tiago

**Additional context**
I am not sure whether to delete Tiago driver? It is not used anymore, but the users may appreciate it as an example.


webots_ros2_tiago/launch/tiago.launch.py
```diff
@@ -35,10 +35,21 @@ def generate_launch_description():
...
+            'wheel_distance': 0.404,
+            'wheel_radius': 0.1955,
+            'left_joint': 'wheel_left_joint',
+            'right_joint': 'wheel_right_joint',
+            'left_encoder': 'wheel_left_joint_sensor',
+            'right_encoder': 'wheel_right_joint_sensor'
+        }],
+        arguments=['--name=tiago_driver'],
+        output='screen'
+    )
```
DavidMansolino, line 52: Shouldn't we remove this file now that it is not used anymore: https://github.com/cyberbotics/webots_ros2/blob/0a4633dbf949abccf076aa9fd91672a901ae2ecf/webots_ros2_tiago/webots_ros2_tiago/tiago_controller.py

lukicdarkoo: I wasn't sure whether to delete it or to make another launch file which uses the old controller. It may be useful to users?

DavidMansolino: As the other node does not provide any advantage for the user (no more feature or so), the only reason for keeping it would be that we may use it as a base to extend it for a more extended support of the Tiago robot. But since the current node is extremely simple it will be very easy to re-create it, in my point of view it is therefore better to remove it to avoid possibly confusing the user.

PR #78, opened 2020-04-28
Extend Reusable Plugins

lukicdarkoo:
**Description**
Webots devices should be automatically discovered and corresponding ROS2 topics/params/services created. This is another step in quick ROSification of robots created in Webots.

**Affected Packages**
  - webots_ros2_*
  - all packages will be affected by this change and all of them should be thoughtfully tested.

**Tasks**
  - [x] Create camera publisher
  - [ ] Create IMU publisher
  - [ ] Create LED subscriber
  - [x] Create distance sensor publisher
    - [x] Integrate lookup table
  - [x] Create light sensor publisher
    - [x] Integrate lookup table
  - [ ] Integrate changes to e-puck:
    - [x] Delete camera publisher
    - [ ] Delete IMU publisher
    - [ ] Delete LED subscriber
    - [x] Delete distance sensor publisher
    - [x] Delete light sensor publisher
    - [ ] Verify the ROS API against documentation
  - [x] Create TF publisher
  - [ ] Convert e-puck robot/world to one compliant to ROS axis convention
    - [ ] Change the documentation accordingly


webots_ros2_core/webots_ros2_core/publishers/publisher.py
```diff
@@ -0,0 +1,6 @@
+class Publisher:
+    def publish(self):
+        pass
```
DavidMansolino, line 3: Probably we can add an assert here to be sure this is overridden in the derived class.

lukicdarkoo:
Peek 2020-05-15 17-26
Here is the current progress of TF publishing. It can be tested with:
```
ros2 launch webots_ros2_tiago tiago.launch.py rviz:=true
```
(I will create a new launch file later)

A few things to note here:
- The system is smart enough to separate static (link-to-link) and dynamic (link-to-joint) transformations, so static transformations are latched topics, published only once, so the resources are not wasted.
- There is a warning for caster wheels (top left corner). This means that the corresponding dynamic transforms are published only at the beginning and there are not updated since then. This is expected because there is no position sensor in the caster wheel joints. We probably should not include those transformations.
- It publishes TFs of all solids with a name attribute configured. This may be a problem because I noticed that two solids can have the same name. Maybe we should constraint this only to devices. But then, the user may not be able to get transformations of specific solids.
- It works properly only with ENU coordinate system. The world included and Tiago use ENU coordinate system and this is the reason why everything is tested with Tiago. With e-puck, the whole robot is rotated in RViz.

lukicdarkoo:
Peek 2020-05-22 11-57
It also works fine with e-puck ENU version (from this PR https://github.com/cyberbotics/webots/pull/1643)

lukicdarkoo:
A different approach is used to implement this in the following PRs: #96, #89, #86 and #83

PR #83, opened 2020-05-18
Create Base for Reusable Plugins 

lukicdarkoo:
**Description**
This PR aims to define a solid base for reusable publishers that will be later extended with more devices.

**Related Issues**
Related to #78 

**Affected Packages**
  - webots_ros2_core
  - webots_ros2_epuck

**Tasks**
  - [x] Camera device
  - [x] LED device
  - [x] Laser device
    - [x] Verify the device
  - [x] Improve lookup table interpolation
    - [x] Add tests
    - [x] Use format from `.getLookupTable()`


lukicdarkoo:
I wasn't sure how to call the "modules". "publisher" is good for publishers, but then there are LEDs which do not publish any data. I settled with "plugin" because [Gazebo](http://gazebosim.org/tutorials?tut=ros_gzplugins) calls them like that, but I am not sure if it is good description either. Let me know if you have a better idea.

omichel:
Why not calling them "devices" instead of "plugins"?

webots_ros2_core/webots_ros2_core/plugins/led_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+
+"""LED plugin."""
+
+from std_msgs.msg import Int32
+from rclpy.qos import qos_profile_sensor_data
+from .plugin import Plugin
+
+
+class LEDPlugin(Plugin):
+    """Webots + ROS2 camera wrapper."""
```
DavidMansolino, line 23:
```suggestion
    """Webots + ROS2 LED wrapper."""
```

webots_ros2_core/webots_ros2_core/plugins/plugin_manager.py
```diff
@@ -0,0 +1,65 @@
...
+try:
+    append_webots_python_lib_to_path()
+    from controller import Node
+except Exception as e:
+    sys.stderr.write('"WEBOTS_HOME" is not correctly set.')
+    raise e
+
+
+class PluginManager:
+    """Publish as ROS topics the laser scans of the lidars."""
```
DavidMansolino, line 32: Need update.

webots_ros2_core/webots_ros2_core/webots_node.py
```diff
@@ -64,6 +64,8 @@ def __init__(self, name, args=None, enableTfPublisher=True):
                 self.get_logger().warn('Impossible to publish transforms because the "supervisor"'
                                        ' field is false.')
 
+        PluginManager(self, plugin_config)
```
DavidMansolino, line 67:
```suggestion
        self.pluginManager = PluginManager(self, plugin_config)
```

I know we are not using it, but looks cleaner to me (and safer tomato sure that the garbage collector doesn't destroy it).

lukicdarkoo:
> Why not calling them "devices" instead of "plugins"?

I was thinking about that but sometimes multiple devices should be combined to get a single topic, e.g. [Imu](https://github.com/ros2/common_interfaces/blob/master/sensor_msgs/msg/Imu.msg) combines InertialUnit, Accelerometer and Gyro. I am not sure if it is going to be the only case.

DavidMansolino:
> > Why not calling them "devices" instead of "plugins"?
> 
> I was thinking about that but sometimes multiple devices should be combined to get a single topic, e.g. [Imu](https://github.com/ros2/common_interfaces/blob/master/sensor_msgs/msg/Imu.msg) combines InertialUnit, Accelerometer and Gyro. I am not sure if it is going to be the only case.

I agree with @omichel, plugins looks like Gazebo, but to me the 'device' term is more clear.
And in the case of IMU, IMU is a device too (even if internally it uses more than one Webots device).

DavidMansolino:
:eyes:

webots_ros2_core/webots_ros2_core/devices/laser_device.py
```diff
@@ -0,0 +1,144 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Laser device."""
+
+from sensor_msgs.msg import LaserScan
+from rclpy.time import Time
+from geometry_msgs.msg import TransformStamped
+from tf2_ros import StaticTransformBroadcaster
```
DavidMansolino, line 20: Unfortunately, this is not available on ROS2 Dashing and before (https://answers.ros.org/question/309889/ros2-modulenotfounderror-no-module-named-tf2_ros/), but we can maybe decide to drop Dashing as supporting the two last version of ROS2 is already good IMO.

lukicdarkoo: `StaticTransformBroadcaster` only defines QoS suitable for static transforms, not hard to avoid using it.

DavidMansolino: If it is easy ok, otherwise (just discussed with @omichel), dropping support for Dashing, is fine (95% of the packages have moved to Eloquent).

webots_ros2_core/CHANGELOG.rst
```diff
@@ -9,6 +9,7 @@ Changelog for package webots_ros2_core
 * Added the possibility to run nodes in synchronized mode (using the 'synchronization' parameter).
 * Added a laser_publisher module.
```
DavidMansolino, line 10: This one can be removed :wink:

webots_ros2_core/webots_ros2_core/devices/led_device.py
```diff
@@ -0,0 +1,46 @@
...
+        # Create publishers
+        self._led_subscriber = node.create_subscription(
+            Int32,
+            self._topic_name,
+            self._callback,
+            qos_profile_sensor_data
+        )
+
+    def _callback(self, msg):
+        self._wb_device.set(msg.data)
```
DavidMansolino, line 43: I am wondering if this should not be a service instead of a topic, is the use of topics a standard way to control led in ROS?

lukicdarkoo: That is a good point. There are two reasons I chose to go with a topic:
- It uses standard message, I didn't have to define my own service and it may be a little easier for users (they don't have that service type on the physical robot, e.g. e-puck in our case).
- There is no feedback from an LED. In the worst case it will not turn on/off, but in general, there is no way to know whether the change is applied.

lukicdarkoo: Ideally, we could read `gradual` and `color` parameters from LED device and use types that are more specific, e.g. Bool, ColorRGBA etc

DavidMansolino: Ok that's fine, I checked and found indeed several examples using topics to control the LEDs.

PR #86, opened 2020-06-04
Utilize Webots URDF Export Function to Publish Transformations

### LINKED ISSUE #76: Improve Tf publisher
**Describe the Bug**
Several improvements can be done to the generic TF publisher (https://github.com/cyberbotics/webots_ros2/blob/develop/webots_ros2_core/webots_ros2_core/tf_publisher.py):
   - TransformBroadcaster should be used instead of a ``tf`` topic for efficiency reasons.
   - For non moving part static transforms should be used, this will require improving this method to check wether the joint are static or dynamic between solids: https://github.com/cyberbotics/webots_ros2/blob/develop/webots_ros2_core/webots_ros2_core/tf_publisher.py#L41
   - The transform should not be published relatively to the ``map`` reference as ``map`` is usually associated with a map frame that has uncertain transformation to robot and it has to be estimate (e.g. particle filter). We should check what is the standard and  what Gazebo is actually doing and rename the reference.

lukicdarkoo:
**Description**
This PR uses new Webots function that allows URDF export (`robot.getUrdf()`) and `robot_state_publisher` to publish transformations.

**Related Issues**
https://github.com/cyberbotics/webots/pull/1706, fixes #76

**Affected Packages**
  - webots_ros2_core
  - webots_ros2_universal_robot
  - webots_ros2_epuck
  - webots_ros2_examples

**Tasks**
  - [x] Add robot device which sets `robot_description` parameter (of `robot_state_publisher` node)
  - [x] Add example usage to `webots_ros2_universal_robot`
  - [x] Replace hard-coded transformations in `webots_ros2_epuck`
  - [x] Add a very basic implementation of universal launcher (used to speed up testing)
  - [x] Delete Navigation2 dependency
  - [x] Add Tiago++ example
  - [x] Documentation
  - [x] Verify impact of RobotDevice on other packages, e.g. Universal Robot (whether there is a clash with `joint_state_publisher` there)
  - [x] Resolve missing textures for e-puck

**Additional context**
In order to keep this PR short a few compromises are made:
- `use_joint_state_publisher` parameter in `webots_node.py` is added temporarily as `JointStatePublisher` should eventually be moved to `/webots_ros2_core/webots_ros2_core/devices/`
- `/webots_ros2_core/launch/robot_launch.py` implements very basic functionalities (some of them in the wrong way, e.g. `world` parameter) and it should be extended in separated PR
- Documentation is briefly introduced and will be extended once more devices are ROSified. At this point, it doesn't make much sense to write proper documentation.

**Useful commands**
```
ros2 launch webots_ros2_universal_robot universal_robot_rviz_dynamic.launch.py
ros2 launch webots_ros2_epuck robot_with_tools_launch.py rviz:=true
ros2 launch webots_ros2_core robot_launch.py world:=$(ros2 pkg prefix webots_ros2_universal_robot --share)/worlds/universal_robot_rviz.wbt
```


wangwangwsf:
How to control the movement of pioneer 3DX in webots?

DavidMansolino:
:eyes:

webots_ros2_core/README.md
```diff
@@ -0,0 +1,127 @@
+# webots_ros2_core
```
DavidMansolino, line 1: Note that some information are already available on the wiki for the released part, it would be nice to integrate them here too so that when we release the new version of the package we can simply copy paste and overwritten the current content of the wiki page: http://wiki.ros.org/webots_ros2/Tutorials/webots_ros2_core

webots_ros2_core/README.md
```diff
@@ -0,0 +1,127 @@
...
+        self.sensor_publisher.publish(msg)
+```
+
+
+### Examples
+This Github repository contains a few good examples that you can use as the starting point:
+- `webots_ros2_example` includes a very simple controller for Thymio (differential driver robot).
+- `webots_ros2_tiago` is another differential drive robot, but here `WebotsDifferentialDriveNode` class from `webots_ros2_core` is utilized to simplify differential drive implementation.
+- `webots_ros2_epuck` is again differential drive robot that creates ROS services and topics for almost all available sensors and actuators.
+This example also contains a [list of instructions](https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_epuck/EPUCK_ROS2.md) that explains how the simulation can be used in combination with different ROS2 packages like RViz and Navigation2.
```
DavidMansolino, line 126:
```suggestion

This example also contains a [list of instructions](https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_epuck/EPUCK_ROS2.md) that explains how the simulation can be used in combination with different ROS2 packages like RViz and Navigation2.
```
after a list an empty line is required.

DavidMansolino: Also I never saw `RViz` written with only the R and V as capital, the following seems more commun:
  - rviz
  - RVIZ
  - (Rviz).

webots_ros2_core/launch/robot_launch.py
```diff
@@ -0,0 +1,85 @@
...
+        print('No world parameter has been found, please define it!')
+        sys.exit(1)
+
+    if not os.path.exists(params['world']):
+        print('World doesn\'t exist')
+        sys.exit(2)
+
+    # Webots
+    arguments = [
+        '--mode=realtime',
```
DavidMansolino, line 43: that could be an optional argument coming from the params.

webots_ros2_core/launch/robot_launch.py
```diff
@@ -0,0 +1,85 @@
...
+        parameters=[{
+            'synchronization': synchronization,
+            'use_joint_state_publisher': True
+        }],
+        output='screen'
+    )
+
+    # Robot state publisher
+    initial_robot_description = '<?xml version="1.0"?><robot name="dummy"><link name="base_link"></link></robot>'
+    rsp = Node(
```
DavidMansolino, line 66: Try avoiding using abbreviations.

webots_ros2_core/webots_ros2_core/devices/robot_device.py
```diff
@@ -0,0 +1,61 @@
...
+
+        # Create robot_description publishers if needed
+        if self._publish_robot_description:
+            urdf = self._wb_device.getUrdf(self.__get_urdf_prefix())
+            self.__save_urdf_to_file(urdf)
+            self.__set_string_param('robot_state_publisher', 'robot_description', urdf)
+
+    def __save_urdf_to_file(self, urdf):
+        """Write URDF to a file for debugging purposes."""
+        with open(os.path.join(tempfile.gettempdir(), 'webots_robot.urdf'), 'w') as urdf_file:
```
DavidMansolino, line 45: This will not work with more than one robot.

webots_ros2_core/webots_ros2_core/devices/robot_device.py
```diff
@@ -0,0 +1,61 @@
...
+import os
+import tempfile
+from .device import Device
+from rcl_interfaces.srv import SetParameters
+from rcl_interfaces.msg._parameter import Parameter
+from rclpy.parameter import ParameterType, ParameterValue
+
+
+class RobotDevice(Device):
+    """Webots + ROS2 Robot wrapper."""
```
DavidMansolino, line 26: Comment to update.

webots_ros2_core/webots_ros2_core/devices/robot_device.py
```diff
@@ -0,0 +1,61 @@
...
+
+import os
+import tempfile
+from .device import Device
+from rcl_interfaces.srv import SetParameters
+from rcl_interfaces.msg._parameter import Parameter
+from rclpy.parameter import ParameterType, ParameterValue
+
+
+class RobotDevice(Device):
```
DavidMansolino, line 25: Even if inherits from `Device` I would not name the class like this, but rather something that really explains what the class is doing, e.g. `RobotDescriptionPublisher`.

webots_ros2_epuck/launch/robot_launch.py
```diff
@@ -50,9 +50,19 @@ def generate_launch_description():
         output='screen'
     )
 
+    # Robot state publisher
+    initial_robot_description = '<?xml version="1.0"?><robot name="dummy"><link name="base_link"></link></robot>'
+    rsp = Node(
```
DavidMansolino, line 55: abbreviation.

webots_ros2_epuck/launch/robot_tools_launch.py
```diff
@@ -22,11 +22,12 @@
 from launch.actions import IncludeLaunchDescription
 from launch.substitutions import LaunchConfiguration
 from launch_ros.actions import Node
-from ament_index_python.packages import get_package_share_directory
+from ament_index_python.packages import get_package_share_directory, get_packages_with_prefixes
 from launch.launch_description_sources import PythonLaunchDescriptionSource
 
 
 def generate_launch_description():
+    ld_nodes = []
```
DavidMansolino, line 30: what `ld` stands for?

webots_ros2_epuck/worlds/epuck_world.wbt
```diff
@@ -26,15 +26,13 @@ WoodenBox {
...
-E-puck {
+E-puck_enu {
   controller "<extern>"
   version "2"
   camera_width 640
   camera_height 480
-  turretSlot [
-    Pi-puck {
-    }
-  ]
```
DavidMansolino, line 37: I assume you removed the extension because the Pi-puck extension id not enu compatible?

webots_ros2_examples/setup.py
```diff
@@ -6,8 +6,10 @@
 data_files = []
 data_files.append(('share/ament_index/resource_index/packages', ['resource/' + package_name]))
 data_files.append(('share/' + package_name, ['launch/example.launch.py']))
-data_files.append(('share/' + package_name + '/worlds',
-                   ['worlds/ros_example.wbt', 'worlds/.ros_example.wbproj']))
+data_files.append(('share/' + package_name + '/worlds', [
+    'worlds/ros_example.wbt', 'worlds/.ros_example.wbproj',
+    'worlds/tiago++_example.wbt', 'worlds/.tiago++_example.wbproj'
```
DavidMansolino, line 11: it's a bit strange to have a tiago example in this package as we have a dedicated tiago package.

webots_ros2_universal_robot/setup.py
```diff
@@ -31,7 +32,7 @@
 data_files.append(('share/' + package_name + '/worlds', worlds))
 data_files.append(('share/' + package_name + '/worlds/textures', textures))
 data_files.append(('share/' + package_name, ['package.xml']))
-
+data_files.append(('share/' + package_name + '/resource', ['resource/view_robot_dynamic.rviz']))
 
 setup(
     name=package_name,
```
DavidMansolino, line 38: It would be nice to create a README.md file for this package (including the information from http://wiki.ros.org/webots_ros2/Tutorials/webots_ros2_universal_robot) to explain the difference between the two rviz examples.

webots_ros2_core/README.md
lukicdarkoo: That capitalization is a huge mess, just on a single page of [the official `rviz` tutorial](http://wiki.ros.org/rviz) they use:
RViz, Rviz and rviz (section `Tutorials`)

In my defence, they use RViz in the logo :)

omichel: image
Yes, this is a strong argument for "RViz" indeed...

webots_ros2_core/webots_ros2_core/devices/robot_device.py
lukicdarkoo: The idea is to further extend this class later with new functionalities, e.g. battery status.

webots_ros2_examples/setup.py
lukicdarkoo: It makes sense, I will move it

webots_ros2_epuck/launch/robot_tools_launch.py
lukicdarkoo: `launch_description`, I will change it

webots_ros2_epuck/worlds/epuck_world.wbt
lukicdarkoo: Yes. It is not a problem changing it, but we already have a lot of code in two places (e-pucks are in `webots` and `webots_ros2`). But if we agree to copy necessary models (converted to ENU) here I will return Pi-puck.

webots_ros2_core/README.md
DavidMansolino: Indeed, then keep it like this :smile:

webots_ros2_core/webots_ros2_core/devices/robot_device.py
DavidMansolino: ok, in that case we can leave it as is but explain clearly what it currently does in the class description.

webots_ros2_epuck/worlds/epuck_world.wbt
DavidMansolino: Maybe we can fix this in a more general way just by opening the `translation` and `rotation` fields of the `Pi-puck` PROTO.

omichel: I would prefer not opening these fields just for fixing a temporary problem. Rather I believe it's better to fix the Pi-puck PROTO on this PR (and in the upcoming FLU oriented e-puck).

DavidMansolino: Honestly, opening a field looks less a hack than duplicating the PROTO to me.

omichel: I don't like opening these fields, because that means there is a possibility to move the Pi-puck card on top the e-puck robot and it will still work, which is not true with the real robot. But we can implement it temporarily in this PR to fix the problem if you want. However, when we publish the FLU oriented e-puck, I would like to remove these fields.

DavidMansolino: If it is temporarily, then I agree it is better to do correct since the begining and go for FLU.

webots_ros2_core/README.md
lukicdarkoo: So, idea is to copy documentation from wiki.ros.org now, keep improving it here and periodically sync the changes with wiki.ros.org?

Having documentation on Github would be useful indeed since index.ros.org takes the docs from Git repos, e.g.:
https://index.ros.org/p/webots_ros2_epuck/github-cyberbotics-webots_ros2/#foxy

DavidMansolino: Yes that sounds to be a good workflow especially for now that several packages are not released.

webots_ros2_epuck/worlds/epuck_world.wbt
lukicdarkoo: I've just checked and it works fine without conversion, textures and images were missing though.

DavidMansolino:
:eyes:

webots_ros2_core/README.md
```diff
@@ -0,0 +1,162 @@
...
+Also, you will this example useful if you are planning later to go to the real robot as we support [ROS2 driver for the real robot](https://github.com/cyberbotics/epuck_ros2) as well.
```
DavidMansolino, line 162: It seems one word is missing here.

webots_ros2_tiago/worlds/tiago++_example.wbt
```diff
@@ -0,0 +1,56 @@
+#VRML_SIM R2020b utf8
+WorldInfo {
+  info [
+    "This simulation shows a Tiago++ making hello with both arms in an industrial environment. It can also be controlled using the arrows on the keyboard."
```
DavidMansolino, line 4: Description should be updated.

webots_ros2_tiago/worlds/tiago++_example.wbt
```diff
@@ -0,0 +1,56 @@
...
+  ]
+  title "Simulation of PAL Robotics' TIAGo++ robot"
```
DavidMansolino, line 6:
```suggestion
  title "ROS2 Simulation of PAL Robotics' TIAGo++ robot"
```

webots_ros2_universal_robot/launch/universal_robot_rviz_dynamic.launch.py
```diff
@@ -0,0 +1,85 @@
...
+    controller = ControllerLauncher(
+        package='webots_ros2_universal_robot',
+        node_executable='universal_robot',
+        parameters=[{'synchronization': synchronization}],
+        output='screen'
+    )
+
+    # Robot state publisher
+    initial_robot_description = '<?xml version="1.0"?><robot name="dummy"><link name="base_link"></link></robot>'
+    rsp = Node(
```
DavidMansolino, line 56: Abbreviation :wink:

PR #94, opened 2020-06-15
Fix array multiplyed by float

DavidMansolino:
**Description**
The returned value of `round` is a float in Python, this float value is then multiplied by with an array, this is quite dangerous in Python.


webots_ros2_epuck/webots_ros2_epuck/simple_mapper.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, name):
         fill_map_param = self.declare_parameter('fill_map', True)
 
         # Init map related elements
-        self.map = [-1]*MAP_WIDTH*MAP_HEIGHT
+        self.map = [-1] * int(MAP_WIDTH) * int(MAP_HEIGHT)
```
omichel: Isn't better to cast it at the origin instead?, e.g.:
```python
MAP_WIDTH = int(round(WORLD_WIDTH / RESOLUTION))
MAP_HEIGHT = int(round(WORLD_HEIGHT / RESOLUTION))
```
Maybe we can even get rid of the the `round` call?

DavidMansolino: Its probably working too, but the result is not 100% the same:
```Python
print(round(5.76543))
print(int(5.76543))
```
=>
```
6
5
```

PR #96, opened 2020-06-16
Add Imu ROS2 Wrapper

lukicdarkoo:
**Description**
This PR wraps Webots' Accelerometer, Gyro and InertialUnit to create Imu topic.

**Affected Packages**
  - webots_ros2_core
  - webots_ros2_epuck

**Tasks**
  - [x] Create mechanism which allows a combination of multiple Webots devices into a single ROS topic
  - [x] Add Imu wrapper
  - [x] Verify the Imu axes

### DIFF
--- a/webots_ros2_core/webots_ros2_core/devices/device_manager.py
+++ b/webots_ros2_core/webots_ros2_core/devices/device_manager.py
@@ -82,7 +82,7 @@ class DeviceManager:
         for device_name in self.__config.keys():
             if device_name not in ['@auto'] and device_name not in self.__devices.keys():
                 self.__node.get_logger().warn(
-                    f'Device `{device_name}` is has not considered! The device doesn\'t exist or it is not supported.')
+                    f'Device `{device_name}` is not considered! The device doesn\'t exist or it is not supported.')
 
         # Create a loop
         self.__node.create_timer(1e-3 * int(node.robot.getBasicTimeStep()), self.__callback)
--- a/webots_ros2_core/webots_ros2_core/devices/device_manager.py
+++ b/webots_ros2_core/webots_ros2_core/devices/device_manager.py
@@ -133,7 +133,7 @@ class DeviceManager:
             elif wb_device.getNodeType() == Node.GYRO:
                 gyros.append(wb_device)
 
-        # If there is only one return the key
+        # If there is only one candiate for Imu create a device and insert it to `self.__devices`
         if len(accelerometers) <= 1 and len(inertial_units) <= 1 and len(gyros) <= 1 and \
                 (len(accelerometers) + len(inertial_units) + len(gyros)) > 0:
             imu_wb_devices = [
--- a/webots_ros2_core/webots_ros2_core/devices/device_manager.py
+++ b/webots_ros2_core/webots_ros2_core/devices/device_manager.py
@@ -23,6 +23,7 @@ from .laser_device import LaserDevice
 from .distance_sensor_device import DistanceSensorDevice
 from .light_sensor_device import LightSensorDevice
 from .robot_device import RobotDevice
+from .imu_device import ImuDevice
 from webots_ros2_core.utils import append_webots_python_lib_to_path
 try:
     append_webots_python_lib_to_path()
@@ -36,9 +37,10 @@ class DeviceManager:
     """Discovers Webots devices and creates corresponding ROS2 topics/services."""
 
     def __init__(self, node, config=None):
-        self._node = node
-        self._devices = {}
-        config = config or {}
+        self.__node = node
+        self.__devices = {}
+        self.__config = config or {}
+        self.__wb_devices = {}
 
         # Determine default global parameters
         self._auto = config.setdefault('@auto', True)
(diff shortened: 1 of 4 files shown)


webots_ros2_core/webots_ros2_core/devices/imu_device.py
```diff
@@ -0,0 +1,94 @@
...
+        self.__inertial_unit = wb_device[2]
+
+        # Create topics
+        self._publisher = None
+        if not self._disable:
+            self._publisher = self._node.create_publisher(Imu, self._topic_name, 1)
+
+    def __enable_imu(self):
+        for wb_device in self._wb_device:
+            if wb_device:
```
DavidMansolino, line 59:
```suggestion
            if wb_device is not None:
```
I have a slight preference for this way, as it clearly expressed what case we are trying to avoid.

webots_ros2_core/webots_ros2_core/devices/device_manager.py
```diff
@@ -36,9 +37,10 @@ class DeviceManager:
     """Discovers Webots devices and creates corresponding ROS2 topics/services."""
 
     def __init__(self, node, config=None):
-        self._node = node
-        self._devices = {}
-        config = config or {}
+        self.__node = node
+        self.__devices = {}
+        self.__config = config or {}
+        self.__wb_devices = {}
```
DavidMansolino, line 43: Why are you using sometimes single underscore and sometimes double underscores?

webots_ros2_core/webots_ros2_core/devices/device_manager.py
```diff
@@ -48,29 +50,100 @@ def __init__(self, node, config=None):
...
-                self._devices[wb_device.getName()] = DistanceSensorDevice(
-                    node, wb_device, config.get(wb_device.getName(), None))
+                device = DistanceSensorDevice(node, wb_device, self.__config.get(wb_device.getName(), None))
             elif wb_device.getNodeType() == Node.LIGHT_SENSOR:
-                self._devices[wb_device.getName()] = LightSensorDevice(node, wb_device, config.get(wb_device.getName(), None))
+                device = LightSensorDevice(node, wb_device, self.__config.get(wb_device.getName(), None))
+
+            # Add device to the list
+            self.__wb_devices[wb_device.getName()] = wb_device
+            if device:
```
DavidMansolino, line 72:
```suggestion
            if device is not None:
```

webots_ros2_core/webots_ros2_core/devices/device_manager.py
```diff
@@ -48,29 +50,100 @@ def __init__(self, node, config=None):
...
+                self.__devices[config_key] = ImuDevice(node, self.__get_imu_wb_devices_from_key(config_key))
 
         # Verify parameters
-        for device_name in config.keys():
-            if device_name not in self._devices and device_name not in ['@auto']:
-                self._node.get_logger().warn(f'There is no device with name `{device_name}`')
+        for device_name in self.__config.keys():
+            if device_name not in ['@auto'] and device_name not in self.__devices.keys():
+                self.__node.get_logger().warn(
+                    f'Device `{device_name}` is has not considered! The device doesn\'t exist or it is not supported.')
```
DavidMansolino, line 85:
```suggestion
                    f'Device `{device_name}` is not considered! The device doesn\'t exist or it is not supported.')
```

webots_ros2_core/webots_ros2_core/devices/device_manager.py
```diff
@@ -48,29 +50,100 @@ def __init__(self, node, config=None):
...
+                    accelerometer = self.__wb_devices[wb_device_name]
+                elif self.__wb_devices[wb_device_name] == Node.INERTIAL_UNIT:
+                    inertial_unit = self.__wb_devices[wb_device_name]
+                elif self.__wb_devices[wb_device_name] == Node.GYRO:
+                    gyro = self.__wb_devices[wb_device_name]
+
+        return [accelerometer, gyro, inertial_unit]
+
+    def __insert_imu_device(self):
+        """Insert Imu device only if non is configured and there is only one in the robot."""
```
DavidMansolino, line 116: if non is configured ? do you mean none?

webots_ros2_core/webots_ros2_core/devices/device_manager.py
lukicdarkoo: I use single underscore for "protected" members and double for "private" members. Although, I am not sure if it is a good convention.

DavidMansolino: looks indeed like a pretty standard convention, thank you I didn't knew about this subtility.

webots_ros2_core/webots_ros2_core/devices/device_manager.py
lukicdarkoo: I've just improved it, maybe it is more clear now. Basically, there are two conditions that have to be fulfilled before automatic device merging (of Accelerometer, Gyro and InertialUnit) is initiated:
- User didn't define any configuration regarding Imu and
- There is no more than one instance of Accelerometer, Gyro or InertialUnit.

Those conditions are fulfilled for the majority of the robots, but if they are not fulfilled it means that the user expects a custom behaviour.

It was a bit tricky to design this, but I am happy to change it if there is a better idea :)

DavidMansolino: > It was a bit tricky to design this, but I am happy to change it if there is a better idea

Yes, I believe you, this is probably the most complex sensor.
About the multiple Accelerometer, Gyro and InertialUnit case, we could use the position extracted from the URDF to group them correctly together, but the current solution is already very good and works perfectly for most of the cases as you said, so let's keep it like this and extend it in the future if we need.

PR #101, opened 2020-07-03
Add bloom-release workflow

DavidMansolino:
**Description**
Automatic bloom-release on release creation.


DavidMansolino:
Unfortunately, this action does not support branch specific version (https://github.com/at-wat/bloom-release-action/blob/master/entrypoint.sh).
But we might create a PR for adding this feature (should be quite simple).

status: on hold for now.

PR #109, opened 2020-07-03
Set Bloom release Package

DavidMansolino:
**Description**

The upmost package is `webots_ros2_desktop` and not the default one )from the repo name `webots_ros2`).


DavidMansolino:
It is still failing because for some reason it is still pointing to `https://github.com/omichel/webots_ros2.git` instead of `https://github.com/cyberbotics/webots_ros2.git` and therefore is not seeing the new 0.0.4 tag.

DavidMansolino:
It looks now correctly into 'https://github.com/cyberbotics/webots_ros2.git' but still refers to 0.0.3

DavidMansolino:
It is definitely using Release 0.0.3 for creating the release :-/
I have the same issue when creating manually the release on my computer, I have to explicitly update the version (which is not possible from the Github action).

DavidMansolino:
Since this workflow is not working and can be confusing, it is probably better to remove it.

lukicdarkoo (approved):
Thanks!

PR #120, opened 2020-08-06
Various macOS fixes

DavidMansolino:
**Description**
This PR fixes various issues with the packages on macOS.


DavidMansolino:
I can't manage to run the ROS2 nodes on macOS (ros2 installation on macOS completely broke my development environment), but now I can at least compile all the packages.

Issue #72, opened 2020-04-23
Generic Robotic Arm node


DavidMansolino:
Similarly to what have been done for differential-wheel robot (https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_core/webots_ros2_core/webots_differential_drive_node.py) we should create a generic class for robotic arms.
Indeed the two currently supported robotic arms share a lot in commun:
  - https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_abb/webots_ros2_abb/abb_driver.py
  - https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_universal_robot/webots_ros2_universal_robot/universal_robot.py

This new class should also provide a generic ros node (e.g. `arm_driver`) that will be compatible with any robotic arms.

PR #124, opened 2020-08-17
Robotic arm node

DavidMansolino:
Implements  #72


webots_ros2_core/README.md
```diff
@@ -209,6 +209,14 @@ ros2 launch webots_ros2_core robot_launch.py \
 ```
 Now, topics `/odom` and `/cmd` should be availabe, so you can read odometry data (e.g. visualize in RViz) and control the robot (with e.g. `teleop_twist_keyboard`).
 
+##### Robotic Arm
+For robotic arm robots you should utilize `webots_robotic_arm_node` which exposes the following parameters:
+```python
+prefix  # prefix to be used for joints
+```
+
+This node will automatically publish the joint state and create an action server to perform trajectory following.
```
omichel, line 218:
```suggestion
This node will automatically publish the joints state and create an action server to perform trajectory following.
```

omichel (changes requested):

omichel (approved):
Thank you.

PR #129, opened 2020-08-19
Runtime Test

DavidMansolino:
**Description**
This PR adds a runtime test, this will make the test more pertinent (before only compilation and coding style was tested).
This will also make it simpler to test the runtime on several versions of ROS2.


lukicdarkoo:
Implemented in #268

PR #131, opened 2020-08-28
Use motor name for Joint_states_publisher

Simon-Steinmann:
Use the name of the motor instead the name of the positionSensor as the published joint name. This is more consistent with ROS convention

### DIFF
--- a/webots_ros2_core/webots_ros2_core/joint_state_publisher.py
+++ b/webots_ros2_core/webots_ros2_core/joint_state_publisher.py
@@ -47,7 +47,7 @@ class JointStatePublisher():
             device = robot.getDeviceByIndex(i)
             if device.getNodeType() == Node.POSITION_SENSOR:
                 motor = device.getMotor()
-                name = motor.getName()
+                name = motor.getName() if motor is not None else device.getName()
                 self.jointNames.append(name)
                 self.sensors.append(device)
                 self.previousPosition.append(0)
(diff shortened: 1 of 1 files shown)


webots_ros2_core/webots_ros2_core/joint_state_publisher.py
```diff
@@ -46,7 +46,8 @@ def __init__(self, robot, jointPrefix, node):
         for i in range(robot.getNumberOfDevices()):
             device = robot.getDeviceByIndex(i)
             if device.getNodeType() == Node.POSITION_SENSOR:
```
DavidMansolino, line 48:
```suggestion
            if device.getNodeType() == Node.Motor:
```

Simon-Steinmann: We only need the motor name, the rest still needs the sensor. So I think my solution is the simpler one. We could switch to select for motor, but then need to assign the sensor by `sensor = device.getPositionSensor()`

DavidMansolino: Oh ok I understand, in that case, we should probably change the name to:
```Python
name = motor.getName() if motor is not None else name = device.getName()
```
Just to handle the case where there is no motor on this joint.

Simon-Steinmann: sounds good to me, can you add the change?

DavidMansolino (approved):
That looks now all good, I just changed the target branch as master should work with the latest official release of Webots.

lukicdarkoo (approved):
Works good, thanks!

PR #139, opened 2020-09-11
Use wget to get Webots instead of urllib.

DavidMansolino:
**Description**
Currently the download of Webots with urlib is failing on the ros buildfarm, I found one package (`fzi_icl_can`) that is using wget successfully on the buildfarm.

It is an old package and a different version of ROS (and therefore the ROS1 buildfarm and not ROS2), but it is worth
 trying.


omichel (approved):
🤞

PR #141, opened 2020-09-15
Remove webots_ros2_desktop

DavidMansolino:
**Description**
Since the compilation of `webots_ros2_desktop`  package is failing on the ros buildfarm,  this PR removes this package and instead adds a mechanism installing Webots on the first use.

See related question: https://answers.ros.org/question/361276/archive-download-failure-from-buildfarm/


DavidMansolino:
Now, if Webots is not installed (or not the right version), when launching the node, a pop-up window will ask the user if he wants to install Webots, if he says yes, the tarball is downloaded and installed in `$HOME/.ros/webots$VERSION`, I have checked but did not found any recommended location to save such files (some other packages such RVIz2 are creating their own hidden folder in the home directory).

webots_ros2_core/webots_ros2_core/webots_launcher.py
```diff
@@ -30,9 +37,52 @@ def __init__(self, *, world, gui, mode):
...
+            else:
+                answer = filedialog.askdirectory(
+                    initialdir=os.getcwd(),
+                    title="If Webots is already installed, please select its home installation folder:"
+                )
+                if isinstance(answer, str):
+                    os.environ['WEBOTS_HOME'] == answer
+                    webots_path = get_webots_home()
+                if webots_path is None:
+                    sys.exit('Missing Webots version "%s"' % get_required_webots_version())
```
omichel, line 84:
```suggestion
                    sys.exit('Could not find Webots version %s' % get_required_webots_version())
```

webots_ros2_core/webots_ros2_core/webots_launcher.py
```diff
@@ -30,9 +37,52 @@ def __init__(self, *, world, gui, mode):
...
+                                         get_required_webots_version())
+            if answer:
+                self.install_webots()
+                webots_path = get_webots_home()
+                if webots_path is None:
+                    sys.exit('Failed to install Webots "%s"' % get_required_webots_version())
+            else:
+                answer = filedialog.askdirectory(
+                    initialdir=os.getcwd(),
+                    title="If Webots is already installed, please select its home installation folder:"
```
omichel, line 78: It is not clear to me that if we answer no to the first dialog, we will have the possibility to specify the installation path in a second step. Maybe the first dialog should say something like "Webots R2020b revision 1 was not found in your system PATH." and have two buttons: "Install it" and "Specify path of already installed version" or something similar.

omichel (changes requested):

webots_ros2_core/webots_ros2_core/utils.py
```diff
@@ -28,62 +28,69 @@
...
+        '/usr/local/webots',  # Linux default install
+        '/snap/webots/current/usr/share/webots',  # Linux snap install
+        '/Applications/Webots.app',  # macOS default install
+        'C:\\Program Files\\Webots',  # Windows default install
+        os.getenv('LOCALAPPDATA', '') + '\\Programs\\Webots'  # Windows user install
+    ]
+    for path in pathes:
+        if os.path.isdir(path) and get_webots_version(path) == version:
+            os.environ['WEBOTS_HOME'] = path
+            return path
```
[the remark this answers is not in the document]
DavidMansolino: > In case Webots was installed in a different folder (from the tarball, or in D:\Program Files\Webots, etc.) this will fail.

In those cases, WEBOTS_HOME should already be set (manually from the user as stated in the user guide: https://www.cyberbotics.com/doc/guide/installation-procedure#installing-the-tarball-package or automatically from the installer on Windows).

> Instead of this, why don't we simply call "which webots" to determine which version is installed and going to be used?

Using which is a good idea, I would however not remove the other methods but add it, as the goal here is to find one specific version of Webots (the user may have several of them installed at different locations).

webots_ros2_core/webots_ros2_core/webots_launcher.py
DavidMansolino: > we will have the possibility to specify the installation path in a second step.

Yes, you have. I will improve the description of the first dialog to make this clearer.

DavidMansolino: Changing the labels of the buttons is not so simple as we can't rely anymore on the default message box, but I have to create my own from scratch.

DavidMansolino:
Included wiki update: https://github.com/cyberbotics/webots_ros2/wiki/Home/_compare/f11e31dc5908ea82096b074374329b63f03e0977...909cdb4dd1ff85700f6343f5860c5698f433372a

webots_ros2_core/webots_ros2_core/webots_launcher.py
```diff
@@ -30,9 +72,61 @@ def __init__(self, *, world, gui, mode):
...
+                    sys.exit('Failed to install Webots %s' % get_required_webots_version())
+            elif dialog.installationMode == InstallationMode.manual:
+                import webbrowser
+                webbrowser.open('https://github.com/cyberbotics/webots/releases/tag/%s' % get_required_webots_version_short())
+                sys.exit()
+            elif dialog.installationMode == InstallationMode.path:
+                from tkinter import filedialog
+                answer = filedialog.askdirectory(
+                    initialdir=os.getcwd(),
+                    title='Please select the folder where Webots is installed:'
```
omichel, line 120:
```suggestion
                    title='Please select the folder where Webots %s is installed:' % get_required_webots_version()
```

omichel (approved):
I may be better to mention the specific version in the dialog for selecting the path.

omichel (approved):
Thank you.

PR #151, opened 2020-10-06
Fix Tests

DavidMansolino:
**Description**
Skipping the submodules test is not a good solution as the tests are run anyway on the buildfarm and are failing.


omichel (approved):
Thank you.

PR #160, opened 2020-10-12
Upgrade PointCloud to PointCloud2

lukicdarkoo:
**Description**
Here we try to utilize new `getPointCloud(dtype='buffer')` (https://github.com/cyberbotics/webots/pull/2321) to increase performance of the LiDAR.

**Related Issues**
- https://github.com/cyberbotics/webots/issues/2283
- https://answers.ros.org/question/362928/publishreceive-large-pointcloud2-is-very-slow/

**Affected Packages**
  - webots_ros2_core

**Tasks**
  - [x] Add to changelog

**Tests**
```bash
ros2 launch webots_ros2_core robot_launch.py \
    executable:=webots_robotic_arm_node \
    world:=$(ros2 pkg prefix webots_ros2_universal_robot --share)/worlds/universal_robot_lidar.wbt
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,6 @@ on:
 
 jobs:
   industrial_ci:
-    if: github.event.pull_request.draft == false
     strategy:
       matrix:
         env:
--- a/webots_ros2_core/CHANGELOG.rst
+++ b/webots_ros2_core/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changelog for package webots_ros2_core
 
 1.0.3 (2020-xx-xx)
 ------------------
-* Improved point cloud publishing by a few times.
+* Improved the performance of point cloud publishing by a few times.
 
 1.0.2 (2020-10-12)
 ------------------
--- a/webots_ros2_core/CHANGELOG.rst
+++ b/webots_ros2_core/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog for package webots_ros2_core
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+1.0.3 (2020-xx-xx)
+------------------
+* Improved point cloud publishing by a few times.
+
 1.0.2 (2020-10-12)
 ------------------
 * Fixed support for 3D Lidars
--- a/webots_ros2_core/webots_ros2_core/devices/lidar_device.py
+++ b/webots_ros2_core/webots_ros2_core/devices/lidar_device.py
@@ -14,7 +14,7 @@
 
 """Lidar device."""
 
-from sensor_msgs.msg import LaserScan, PointCloud
+from sensor_msgs.msg import LaserScan, PointCloud2, PointField
 from geometry_msgs.msg import Point32
 from tf2_ros import StaticTransformBroadcaster
 from geometry_msgs.msg import TransformStamped
@@ -30,7 +30,7 @@ class LidarDevice(SensorDevice):
 
     It allows the following functinalities:
     - Publishes range measurements of type `sensor_msgs/LaserScan` if 2D Lidar is present
-    - Publishes range measurements of type `sensor_msgs/PointCloud` if 3D Lidar is present
+    - Publishes range measurements of type `sensor_msgs/PointCloud2` if 3D Lidar is present
 
     Args:
         node (WebotsNode): The ROS2 node.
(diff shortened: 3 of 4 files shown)


lukicdarkoo:
With the current implementation, we get a better result (maybe double of speed). However, this is still slow and we have to make it faster.

I benchmarked the implementation and the bottleneck seems to be in ROS2:
```Python
from time import time
import array
import rclpy
from rclpy.node import Node
from std_msgs.msg import Int8MultiArray


MESSAGE_N_BYTES = int(1e6)
PUBLISH_PERIOD = 0.01


class TestPublisher(Node):
    def __init__(self):
        super().__init__('test_publisher')
        self.publisher = self.create_publisher(Int8MultiArray, 'test', 10)
        self.timer = self.create_timer(PUBLISH_PERIOD, self.timer_callback)
        self.data = [0] * MESSAGE_N_BYTES
        self.msg = Int8MultiArray()
        self.msg.data = self.data
        self.data_array = array.array('b', self.data)

    def timer_callback(self):
        tic = time()
        msg = Int8MultiArray()
        msg_ms = (time() - tic) * 1000

        tic = time()
        # msg.data = self.data_array
        msg.data = self.data
        data_ms = (time() - tic) * 1000

        tic = time()
        self.publisher.publish(self.msg)
        pub_ms = (time() - tic) * 1000

        print('Create message', int(msg_ms))
        print('Fill with data', int(data_ms))
        print('Publish message', int(pub_ms))


def main(args=None):
    rclpy.init(args=args)
    test_publisher = TestPublisher()
    rclpy.spin(test_publisher)
    test_publisher.destroy_node()
    rclpy.shutdown()


if __name__ == '__main__':
    main()
```

Surprisingly, the bottleneck is in the following line:
```
msg.data = self.data
```

as the timings are the following:
```
Fill with data 122
Publish message 10
Create message 0
```

Looking under-the-hood, each message type with the `data` attribute has a setter which converts the input value to Python array:
```Python
    @data.setter
    def data(self, value):
        if isinstance(value, array.array):
            assert value.typecode == 'b', \
                "The 'data' array.array() must have the type code of 'b'"
            self._data = value
            return
        if __debug__:
            from collections.abc import Sequence
            from collections.abc import Set
            from collections import UserList
            from collections import UserString
            assert \
                ((isinstance(value, Sequence) or
                  isinstance(value, Set) or
                  isinstance(value, UserList)) and
                 not isinstance(value, str) and
                 not isinstance(value, UserString) and
                 all(isinstance(v, int) for v in value) and
                 all(val >= -128 and val < 128 for val in value)), \
                "The 'data' field must be a set or sequence and each value of type 'int' and each integer in [-128, 127]"
        self._data = array.array('b', value)
```

This conversion (`self._data = array.array('b', value)`) actually takes the most of the time.

In conclusion, we can return the correct array type directly from Webots and in that way avoid the ROS2 bottleneck. Working on that...

.github/workflows/test.yml
```diff
@@ -6,7 +6,6 @@ on:
 
 jobs:
   industrial_ci:
-    if: github.event.pull_request.draft == false
```
omichel, line 9: Is that on purpose? I would assume yes and it's fine, but I just want to be sure you are aware of it.

webots_ros2_core/CHANGELOG.rst
```diff
@@ -2,6 +2,10 @@
 Changelog for package webots_ros2_core
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+1.0.3 (2020-xx-xx)
+------------------
+* Improved point cloud publishing by a few times.
```
omichel, line 7:
```suggestion
* Improved the performance of point cloud publishing by a few times.
```

omichel (commented):
Very good.

.github/workflows/test.yml
lukicdarkoo: Yes. That line doesn't make sense for me, as I want to see if my CI is failing before I request a review. It is already deleted in the master branch (I will master to develop later).

lukicdarkoo:
I fixed the warning from DeepCode

webots_ros2_core/webots_ros2_core/devices/lidar_device.py
```diff
@@ -92,12 +91,28 @@ def step(self):
             self._wb_device.disable()
 
     def __publish_point_cloud_data(self, stamp):
-        points = self._wb_device.getPointCloud()
-        if points:
-            msg = PointCloud()
+        data = self._wb_device.getPointCloud(dtype='buffer')
```
omichel, line 94:
```suggestion
        data = self._wb_device.getPointCloud(data_type='buffer')
```

omichel (changes requested):

webots_ros2_core/webots_ros2_core/devices/lidar_device.py
lukicdarkoo: Thanks!

omichel (approved):
Thank you.

PR #162, opened 2020-10-19
Fix webots_ros2_importer

DavidMansolino:
**Description**
This package is having several issues.

### DIFF
--- a/webots_ros2_importer/test/ament_flake8.ini
+++ b/webots_ros2_importer/test/ament_flake8.ini
@@ -1,0 +1,6 @@
+[flake8]
+extend-ignore = C816,D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404,I202,E402
+import-order-style = google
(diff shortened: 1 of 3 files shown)


lukicdarkoo:
It is not the prettiest solution, but it works :)

DavidMansolino:
> It is not the prettiest solution, but it works :)

But pep8 doesn't like it :laughing: 
https://travis-ci.com/github/cyberbotics/webots_ros2/jobs/402037530#L1409

We probably have to ignore `E402` in the pep8 test of this module.

lukicdarkoo:
I didn't know we can ignore errors in comments (e.g. `# noqa: E402`). It's a much better solution than creating a custom config file.

DavidMansolino (commented):
Thank you.
I approve this PR  (I can't formally approve since I am the original author).

lukicdarkoo (approved):
I approved it :)

PR #166, opened 2020-11-09
Increase version number

lukicdarkoo:
Develop branch should target R2021a. Also, copyright was wrong in the headers


omichel (approved):
Thank you.

PR #174, opened 2020-12-01
Remove unnecessary joints from the UR5e proto

lukicdarkoo:
While preparing the package for the next release:
https://github.com/cyberbotics/webots_ros2/wiki/Internal-Tests

I spotted there are some errors in the terminal due to extra links (I believe that some of them are added to work with deprecated [TF publisher](https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_core/webots_ros2_core/tf_publisher.py)) in the URDF file when running:
```
ros2 launch webots_ros2_universal_robot universal_robot_rviz.launch.py
```

### DIFF
--- a/webots_ros2_ur_e_description/urdf/ur5e_robot.urdf
+++ b/webots_ros2_ur_e_description/urdf/ur5e_robot.urdf
@@ -225,19 +225,6 @@
       <inertia ixx="9.890410052167731e-05" ixy="0.0" ixz="0.0" iyy="9.890410052167731e-05" iyz="0.0" izz="0.0001321171875"/>
     </inertial>
   </link>
-  <joint name="ee_fixed_joint" type="fixed">
-    <parent link="wrist_3_link"/>
-    <child link="ee_link"/>
-    <origin rpy="0.0 0.0 1.5707963267948966" xyz="0.0 0.1 0.0"/>
-  </joint>
-  <link name="ee_link">
-    <collision>
-      <geometry>
-        <box size="0.01 0.01 0.01"/>
-      </geometry>
-      <origin rpy="0 0 0" xyz="-0.01 0 0"/>
-    </collision>
-  </link>
   <transmission name="shoulder_pan_trans">
     <type>transmission_interface/SimpleTransmission</type>
     <joint name="shoulder_pan_joint">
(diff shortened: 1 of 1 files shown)


omichel:
Shall we modify these files? It seems they are [automatically generated](https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_ur_e_description/urdf/ur5e_robot.urdf#L2-L5). Also, the UR3 and UR10 models seem to have the same problem, isn't it?

lukicdarkoo:
> Shall we modify these files? It seems they are [automatically generated](https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_ur_e_description/urdf/ur5e_robot.urdf#L2-L5). Also, the UR3 and UR10 models seem to have the same problem, isn't it?

I think the files are generated from XACRO files before they are deployed to the repository. It is true that UR3 and UR10 also have the same problem but we don't have any demo for them.

I believe a bigger problem is that the PROTO and URDF files are independent. Therefore, changing the PROTO file will not affect the URDF file and vice versa. This can lead to unexpected behavior, e.g. changing a link name in the PROTO file can will break the RVIz2 visualization. I would like to eventually delete the `webots_ros2_ur_e_description` package and to use only the Webots' URDF exporter.

omichel:
I thought the PROTO could be generated automatically from the URDF. Isn't it the case here?
My feeling is that the global picture is this: `XACRO` &rarr; `URDF` &rarr; `PROTO`. So, if the `PROTO` or `URDF` needs to be fixed, we should fix the `XACRO` file and re-generate the `URDF` and `PROTO` from it. Do you agree?

lukicdarkoo:
> I thought the `PROTO` could be generated automatically from the `URDF`. Isn't it the case here?
> My feeling is that the global picture is this: `XACRO` → `URDF` → `PROTO`. So, if the `PROTO` or `URDF` needs to be fixed, we should fix the `XACRO` file and re-generate the `URDF` and `PROTO` from it. Do you agree?

Yes, that is also a good way to go. Probably better approach as it provides added value to the users, it shows `XACRO` → `URDF` → `PROTO` conversion example.

This will take more time though. Do you agree that we create an issue about that, merge this hotfix, so that we can release the next webots_ros2 version before Foxy sync? I believe the next Foxy sync is on Friday, so we should publish the package today or tomorrow (the latest).

omichel:
> This will take more time though. Do you agree that we create an issue about that, merge this hotfix, so that we can release the next webots_ros2 version before Foxy sync? I believe the next Foxy sync is on Friday, so we should publish the package today or tomorrow (the latest).

Sure.

omichel (approved):

DavidMansolino:
> My feeling is that the global picture is this: XACRO → URDF → PROTO.

That's exactly how it was done. To be more precise, the XACRO was taken from the 'ur_e_description' package (https://github.com/ros-industrial/universal_robot/tree/kinetic-devel/ur_e_description, not yet available in ROS2, but once available in ROS2 it should be used instead and removed from here) and the URDF was committed instead of the XACRO because at that time, XACRO was not available in ROS2.
I believe the resulting URDF should stay as close as possible to the original one as it is the 'official' ROS one (and if the examples are not compatible with it, it is probably better to improve the examples than changing the URDF file).

> This can lead to unexpected behavior, e.g. changing a link name in the PROTO file can will break the RVIz2 visualization.

This should be avoided at any cost, as those names are the 'official' ones of the URe robots in ROS/ROS2, changing them might break compatibility with many packages related to these robots.

PR #191, opened 2021-01-05
Fix CI

lukicdarkoo:
After the previous PR, the CI started reporting `Matrix options must only contain primitive values`:
image
This looks like GitHub's internal change.

This PR fixes the problem as well as a few minor issues.

----

The required check `industrial_ci (foxy, main)` should be replaced by `CI / industrial_ci (main, foxy) (push)`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,22 +1,19 @@
 name: CI
 
-on: 
-  pull_request:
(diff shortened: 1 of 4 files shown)


omichel (approved):

omichel (approved):
Great.

PR #200, opened 2021-02-16
add recognition publisher and message

renan028:
**Description**
Feature: enable camera recognition, available in Webots.
It checks if the camera node has the "recognition" field and it starts publishing a custom message (something similar to webots_ros1).

**Affected Packages**
List of affected packages:
  - webots_ros2_core
  - webots_ros2_msgs


lukicdarkoo:
Thank you for the contribution!

Can we use standard messages instead?
https://github.com/ros-perception/vision_msgs/tree/ros2/msg

E.g., `BoundingBox3DArray.msg` seems like a suitable message type.

renan028:
Unfortunately `BoundingBox3DArray.msg` will not give me all the info available.
I thought of using `Detection2DArray.msg` so I could add the object's model. But then "`colors`" is missing and some other info: "`position in image`", "`size in image`" (that I do not care much about, but I think it should be provided in the API).
What do you think?

lukicdarkoo:
Ok, I see. Let's publish two topics then:
```
/{camera_name}/recognitions [BoundingBox3DArray]
/{camera_name}/recognitions/webots [WbCameraRecognitionObject]
```
@omichel What do you say? We can make `/{topic_name}/webots` a rule for all Webots specific APIs.

@renan028 I couldn't find a nice example of how this is resolved in the community. Do you have any better idea?

omichel:
> @omichel What do you say? We can make /{topic_name}/webots a rule for all Webots specific APIs.

Yes, that seems a good idea. Note that recognition should probably use the singular form (without final s).

renan028:
```
/{camera_name}/recognition [BoundingBox3DArray]
/{camera_name}/recognition/webots [WbCameraRecognitionObject]
```
Fine by me too. `BoundingBox3DArray` could be important for navigation (obstacle avoidance).

> @renan028 I couldn't find a nice example of how this is resolved in the community. Do you have any better idea?
@lukicdarkoo while it’s good practice to use predefined interfaces, since `std_msgs`, `geometry_msgs` and `vision_msgs` do not provide a suitable contract for Webots's object recognition, I think it is better to merge all available info in a single msg than to create several msgs (that one would need to subscribe and link them somehow to get the full picture). It is just important to improve documentation.
However, I think we should use the `geometry_msgs` instead of plain `std_msgs` for `WbCameraRecognitionObject`, i.e., instead of:
```
std_msgs/Header header
float32[3] position
float32[4] orientation
int32[2] position_on_image
int32[2] size_on_image
int32 number_of_colors
float32[] colors
string model 
```
We could use:
```
geometry_msgs/PoseStamped pose
int32[2] position_on_image
int32[2] size_on_image
int32 number_of_colors
std_msgs/ColorRGBA[] colors
string model 
```
What do you think?

lukicdarkoo:
I wanted to write `Detection2DArray` instead of `BoundingBox3DArray` as you suggested, sorry. So:
```
/{camera_name}/recognition [Detection2DArray]
/{camera_name}/recognition/webots [WbCameraRecognitionObject]
```

@renan028 

> I think it is better to merge all available info in a single msg than to create several msgs

The `WbCameraRecognitionObject` message should contain everything that the corresponding `WbCameraRecognitionObject` contains, no missing information. The `Detection2DArray` message should serve the purpose of general compatibility with the other packages.


> ```
> geometry_msgs/PoseStamped pose
> int32[2] position_on_image
> int32[2] size_on_image
> int32 number_of_colors
> std_msgs/ColorRGBA[] colors
> string model 
> ```
> 
> What do you think?

I think it looks much better than the raw version. I would slightly change it:
```
int32 id
geometry_msgs/PoseStamped pose
vision_msgs/BoundingBox2D bbox (position_on_image + size_on_image)
std_msgs/ColorRGBA[] colors (this contains information about the size)
string model
```
It looks much more ROS, but I am a bit worried about documentation. The raw fields are properly documented in the Webots documentation and changing their type would require documenting them again. This is a bit hard to maintain.

@omichel 

> Yes, that seems a good idea. Note that recognition should probably use the singular form (without final s).

There are multiple objects returned per image. The topic should correspond to the `wb_camera_recognition_get_objects` function.

lukicdarkoo:
@renan028 
It looks neat now!

Rolling is reporting some missing packages which fine, we will not distribute them to Rolling until dependencies are published there. However, the Rolling tests are canceling the Foxy tests, so we can merge.

I added some changes to the master branch, can you merge it?

renan028:
Sure, I will do it! 
And there is a fix that I need to do also: add both subscriptions counts 
```
if self._wb_device.hasRecognition() and (
                self._recognition_publisher.get_subscription_count() > 0 or
                self._recognition_webots_publisher.get_subscription_count() > 0):
```
I was testing it and, besides that, everything seems to work fine.

lukicdarkoo:
The missing package in Rolling is not hiding the other errors in CI. There are some minor cosmetic errors (`industrial_ci (main, foxy)`) to fix to keep the CI calm.

lukicdarkoo (approved):
Thank you!

@omichel, please add privileges to @renan028 so he can merge the PR.

omichel:
> @omichel, please add privileges to @renan028 so he can merge the PR.

Invitation sent!

PR #208, opened 2021-02-24
add gps_device

renan028:
**Description**
Add GPS device for webots_ros2 API. 

**Affected Packages**
List of affected packages:
  - webots_ros2_core

**Additional context**
You can test it by adding:
```
GPS {
      accuracy 0.5
      speedNoise 2
    }
```
In the tesla model and creating a subscriber type `PointStamped`, topic `gps/gps`

### DIFF
--- a/webots_ros2_core/webots_ros2_core/devices/gps_device.py
+++ b/webots_ros2_core/webots_ros2_core/devices/gps_device.py
@@ -76,7 +76,9 @@ class GpsDevice(SensorDevice):
         if not stamp:
             return
 
-        if self.__gps_publisher.get_subscription_count() > 0 or self._always_publish:
+        if self.__gps_publisher.get_subscription_count() > 0 or \
+            self.__speed_publisher.get_subscription_count() > 0 or \
+                self._always_publish:
             self._wb_device.enable(self._timestep)
             msg = Float32()
             msg.data = self._wb_device.getSpeed()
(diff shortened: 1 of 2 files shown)


webots_ros2_core/webots_ros2_core/devices/gps_device.py
```diff
@@ -0,0 +1,103 @@
...
+        else:
+            self.__gps_publisher = node.create_publisher(
+                PointStamped, self._topic_name + '/gps', qos_profile_sensor_data)
+
+    def step(self):
+        stamp = super().step()
+        if not stamp:
+            return
+
+        if self.__gps_publisher.get_subscription_count() > 0 or self._always_publish:
```
lukicdarkoo, line 79:
```suggestion
        if self.__gps_publisher.get_subscription_count() > 0 or self.__speed_publisher.get_subscription_count() > 0 or self._always_publish:
```
If a user subscribes to `/speed` and not to `/gps` it should also work

lukicdarkoo (changes requested):
Nice! Make sure that we use a maximum line length of 128, not the default line length.

lukicdarkoo (approved):
Thank you!

We should consider exposing a velocity vector-like here:
http://wiki.ros.org/hector_gazebo_plugins#GazeboRosGps
(this has to be done in the Webots core)

renan028:
Yes! I agree.

Issue #210, opened 2021-03-01
Integrate `ros2_control`


lukicdarkoo:
The [ros2_control](https://discourse.ros.org/t/announcing-ros2-control-for-foxy/18274) package has been released recently and it would useful to integrate it.

Once we have the [ros2_control](https://github.com/ros-controls/ros2_control) integrated we can get rid of:
- `webots_differential_drive_node.py` in favor of [diff_drive_controller](https://github.com/ros-controls/ros2_controllers/tree/master/diff_drive_controller)
- `joint_state_publisher.py` in favor of [joint_state_controller](https://github.com/ros-controls/ros2_controllers/tree/master/joint_state_controller)
- `trajectory_follower.py` in favor of [joint_trajectory_controller](https://github.com/ros-controls/ros2_controllers/tree/master/joint_trajectory_controller)
- `webots_robotic_arm_node.py` will be useless as well.

Each controller coming from the `ros2_controllers` package is better designed + we will get access to [more controllers](https://github.com/ros-controls/ros2_controllers) (there will be even more as they have recently started porting them from ROS 1). Other benefits are:
- the ability to create a custom controller or use an existing custom controller,
- use `ros2controlcli`,
- replace controllers at runtime,
- manage ownership,
- and other perks coming with the [ros2_control](http://www.theoj.org/joss-papers/joss.00456/10.21105.joss.00456.pdf).

The problem is that `ros2_control` provides only a C++ interface. Therefore, there are two ways to integrate the package:
1. Translate `webots_ros2_core/webots_node.py` and `webots_ros2_core/devices` to C++, and create Python bindings for `webots_node.py` so `WebotsNode` can be extended in Python as well.
2. Create a standalone `ros2_control` based node that will exchange metadata about joints with the Python driver. It means that the `read/write` functions would interact with `WebotsNode` through custom ROS2 messages.

Resources:
- [ros2_control docs](https://ros-controls.github.io/control.ros.org/getting_started.html#compiling)
- [ros_control in general](https://fjp.at/posts/ros/ros-control/)

PR #212, opened 2021-03-15
Renan/webots ros2 cpp

renan028:
**Description**
Webots-ROS2 Cpp interface
It requires a lot of sensors and controllers integration yet, but it is a starting point.
It needs an example (launch file with the `wb_task_node`).
And a minimal diff_driver_controller (ros2_control integration?).
Please feel free to improve.

In the core, It adds a launch file to just launch webots simulator without controllers (because that would be a job for the cpp node).

**Affected Packages**
List of affected packages:
  - webots_ros2_core

### DIFF
--- a/webots_ros2_core/launch/webots_launch.py
+++ b/webots_ros2_core/launch/webots_launch.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python
+
+# Copyright 1996-2021 Cyberbotics Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Launch Webots and ROS2 driver."""
+
+import os
+import launch
+from launch import LaunchDescription
+from launch.actions import RegisterEventHandler, EmitEvent, DeclareLaunchArgument
+from launch.substitutions import LaunchConfiguration
+from launch_ros.actions import Node
+from webots_ros2_core.utils import ControllerLauncher
+from webots_ros2_core.webots_launcher import WebotsLauncher
+
+
+ARGUMENTS = [
+    DeclareLaunchArgument(
+        'world',
+        description='Path to Webots\' world file. Make sure `controller` field is set to `<extern>`.'
+    ),
+    DeclareLaunchArgument(
+        'gui',
+        default_value='True',
+        description='Whether to start GUI or not.'
+    ),
+    DeclareLaunchArgument(
+        'mode',
+        default_value='realtime',
+        description='Choose the startup mode (it must be either `pause`, `realtime`, `run` or `fast`).'
+    ),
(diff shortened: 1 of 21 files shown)


lukicdarkoo:
Very nice! Thank you for this initiative!

> And a minimal diff_driver_controller (ros2_control integration?).

Yes, we want to do everything through `ros2_control`. Even sensors such as IMU should be integrated through `ros2_control`.

What is your motivation behind this? We have been considering moving everything to C++ because of `ros2_control` (see https://github.com/cyberbotics/webots_ros2/issues/210). However, Python is a very friendly way to extend a driver. Ideally, we would keep both interfaces (C++ and Python).

This is a very big change (most probably needed one). Therefore, we have to understand challenges quite well and tailor the design carefully.

We should keep the following in mind:
- https://github.com/cyberbotics/webots_ros2/issues/210: As stated in the issue, we want to get rid of the custom-made controllers in favor of the controllers from `ros2_control`.
- https://github.com/cyberbotics/urdf2webots/issues/112: All configurations that are currently done through ROS parameters (e.g. https://github.com/cyberbotics/webots_ros2/blob/d0f772b358252cda6b9e56768e5ec5cf7454a5ff/webots_ros2_core/webots_ros2_core/devices/sensor_device.py#L34-L38) should live in the `<webots>` tag.
- The C++ interface should be easily expandable with C++ and Python.

renan028:
I was the main contributor of the [Miron-Project](https://github.com/MiRON-project), which uses Webots simulator and ROS1 to test some academic research (mainly contextual navigation).
My last contribution to the project was the [tiago_webot_ros](https://github.com/MiRON-project/tiago_webots_ros), a package that integrates the Tiago robot (Pal Robotics) in Webots with the ROS1 Navigation Package.  

The project ended in 2020. In my spare time, I started to learn ROS2 and to integrate Navigation2 (ROS2) on the same [Tiago project](https://github.com/renan028/tiago_webots_ros2). My goal was to accomplish the same results with ROS2. I manage to integrate everything with the webots_ros2 python package, but I am getting huge offsets in localization and problems to set goals sometimes (I think ROS2/Navigation2 is not stable yet or maybe the problem is mixing python with C++ ? I have to debug). 

Currently, I am working with Gazebo at work, but I like to explore other simulators in my spare time. I had a nice experience with Webots before, I like the webots_ros API and I think webots_ros2 is a nice start to play with ROS2. And, I will probably try multi-robot soon on webots ROS1, because I am not happy with Gazebo for this particular purpose. 

I will take a look at the ros2_control and the other links you shared, in my spare time, and share with you my thoughts on this integration. But, as you said, this is a big change and it requires a well thought (software architect) design. This draft is currently only a ROS2-cpp version of some sensors (not IMU, which should be in ros2_control).

lukicdarkoo:
> My last contribution to the project was the tiago_webot_ros, a package that integrates the Tiago robot (Pal Robotics) in Webots with the ROS1 Navigation Package.

Nice project!

> but I am getting huge offsets in localization and problems to set goals sometimes

Maybe the odometry is not correct, like wheel separation or wheel radius, or there is some bug in sampling? Anyhow, `ros2_control` should resolve it.

> I will take a look at the ros2_control and the other links you shared, in my spare time, and share with you my thoughts on this integration. But, as you said, this is a big change and it requires a well thought (software architect) design. 

We can have a video call to discuss the new design :) I will try to specify some design ideas before.

renan028:
> We can have a video call to discuss the new design :) I will try to specify some design ideas before.
That would be great. Feel free to add me to a channel where we can discuss it.

lukicdarkoo:
Implemented in #221

PR #181, opened 2020-12-29
Add MoveIt 2 demo

lukicdarkoo:
This PR adds the MoveIt2 demo.

It also includes the following improvements:
- [x] refactor `trajectory_follower.py`
  - [x] improve performance (from ~0.7x to ~15x) 
  - [x] handle requests that don't specify time
  - [x] send feedback
  - [x] impose PEP conventions
- [x] allow simulation to run faster than real-time (`webots_node.py`)
- [x] impose PEP conventions to `webots_ros2_demos`

PR #213, opened 2021-03-23
Make the MoveIt example use the new URDF CLI tool

lukicdarkoo:
This PR should solve problems listed in https://github.com/cyberbotics/webots_ros2/pull/181

It should be possible using:
https://github.com/cyberbotics/webots/pull/2885


lukicdarkoo:
I am closing this PR as #345 will bring a better way to handle this problem (we will have URDF stored on disk).

PR #219, opened 2021-04-09
Bump the version number

lukicdarkoo:
I am also deleting the Rolling tests as there are missing packages: `vision_msgs` and `ackermann_msgs`. I posted an issue about the missing packages:
- https://github.com/ros-perception/vision_msgs/issues/55
- https://github.com/ros-drivers/ackermann_msgs/issues/14#issuecomment-816511743

As soon as the packages are available I will enable the CI tests for Rolling. Now, the plan is to create a `rolling` branch and delete all our packages that depend on the missing packages.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         ROS_REPO: [main, testing]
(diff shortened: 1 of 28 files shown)


omichel (approved):
That seems to be a good plan.

PR #233, opened 2021-06-09
fixed URLS

BenjaminDeleze:
**Description**
Use a https URL for texture image in Webots R2021b.
Relates to cyberbotics/webots#2787

The urls are set to R2021a because R2021b does not exist yet. This is the same compromise as the one adopted in cyberbotics/webots_ros#59

The change in `E-puckDistanceSensor_enu.proto` correspond to the one present in the Webots to solve this error: `The 'IndexedLineSet' node doesn't support 'PBRAppearance' in the 'appearance' field of its parent node, please use 'Appearance' instead.`

I had to remove the line `data_files.append(('share/' + package_name + '/worlds/textures', ['worlds/textures/ground.jpg']))` from the `setup.py`of webots_ros2_tesla but I am not sur of what to do in the cases of the other packages.
For example, the `setup.py` of webots_ros2_examples contains this instructions: `data_files.append(('share/' + package_name + '/protos/textures', glob('protos/textures/*')))` that is now useless as there is no more textures in the protos/textures of this package.
But, the same line is also present in the `setup.py` of the turtlebot package where there is not textures from the start. So I am not sure if I am supposed to remove this line from all packages or let it.


lukicdarkoo:
> For example, the `setup.py` of webots_ros2_examples contains this instructions: `data_files.append(('share/' + package_name + '/protos/textures', glob('protos/textures/*')))` that is now useless as there is no more textures in the protos/textures of this package.
> But, the same line is also present in the `setup.py` of the turtlebot package where there is not textures from the start. So I am not sure if I am supposed to remove this line from all packages or let it.

Delete it :)

webots_ros2_epuck/protos/E-puck_enu.proto
```diff
@@ -89,7 +89,7 @@ Robot {
                       roughness 0.4
                       baseColorMap ImageTexture {
                         url [
-                          "textures/gctronic_logo.png"
+                          "https://raw.githubusercontent.com/cyberbotics/webots/R2021a/projects/robots/gctronic/e-puck/protos/textures/gctronic_logo.png"
```
lukicdarkoo, line 92: Shouldn't we target a tag or a commit? This way, if we move/rename textures we will break a lot of things

lukicdarkoo: We used the `R2021a` tag in the webots_ros package. That seems to be a good compromise

BenjaminDeleze:
I fixed the URLs that were targeting master instead of R2021a

lukicdarkoo (approved):
Thank you

PR #236, opened 2021-06-21
Add controller manager spawner to the TurtleBot launch file

lukicdarkoo:
The TurtleBot launch file now automatically starts controllers (`diffdrive_controller` and `joint_state_broadcaster`)

The prefix trick is needed until either of these is released (it takes more than 10s to load the simulation, so the spawner fails):
- https://github.com/ros-controls/ros2_control/pull/444
- https://github.com/ros2/launch/pull/514


omichel (approved):
Thank you.

PR #248, opened 2021-07-19
Merge master into foxy

lukicdarkoo:
no conflicts

PR #251, opened 2021-07-19
Fix webots_ros2_control changelog

lukicdarkoo:
Reference: https://github.com/cyberbotics/webots_ros2/pull/248#discussion_r672158036

I would also like to raise a question here. Should the date be related to release or feature merging or tag?


stefaniapedrazzi (approved):
Thank you.

I think it is fine to set the release date to today.
This should then correspond to the tag date and the date we apply final changes on the version in this repository.

PR #255, opened 2021-07-22
Fix warnings

lukicdarkoo:
The change in `Ros2Control.cpp` fixes the following warning:
```
/home/lukic/foxy_ws/src/webots_ros2/webots_ros2_control/src/Ros2Control.cpp:40:110: warning: unused parameter ‘parameters’ [-Wunused-parameter]
```

Some parameters in the `DirectionPanel` are deleted, so we have to delete them in the world file as well.

### DIFF
--- a/webots_ros2_control/src/Ros2Control.cpp
+++ b/webots_ros2_control/src/Ros2Control.cpp
@@ -37,10 +37,9 @@ namespace webots_ros2_control
     mControllerManager->write();
   }
 
-  void Ros2Control::init(webots_ros2_driver::WebotsNode *node, std::unordered_map<std::string, std::string> &parameters)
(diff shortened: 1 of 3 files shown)


webots_ros2_control/src/Ros2Control.cpp
```diff
@@ -40,6 +40,7 @@ namespace webots_ros2_control
   void Ros2Control::init(webots_ros2_driver::WebotsNode *node, std::unordered_map<std::string, std::string> &parameters)
```
Timple, line 40: I think you can simply remove the word `parameters` on this line (leave the type definition).

BenjaminDeleze (approved):
Seems good for me

Issue #239, opened 2021-06-22
Consider deleting some packages


lukicdarkoo:
We have a lot of packages, some of them don't provide added value, so I would consider deleting them.

| Simulation                   | Proposed Action | Comment                                                                                                                                                                     |
|------------------------------|-----------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| webots_ros2_abb              | delete          | No added value, just a dummy simulation                                                                                                                                     |
| webots_ros2_demos            | rename          | It demonstrates how to use multiple robots. Alternatively, the multi-robot simulations can be demonstrated with UR5 robots (already exists in webots_ros2_universal_robot)  |
| webots_ros2_core             | deprecate       | After the webots_ros2_driver is mature the webots_ros2_core should be deprecated                                                                                            |
| webots_ros2_examples         | delete          | No added value                                                                                                                                                              |
| webots_ros2_tiago            | delete          | No added value                                                                                                                                                              |
| webots_ros2_tutorials        | move            | It should be maintained by SoftIllusion                                                                                                                                     |
| webots_ros2_universal_robot  | cleanup        | Very useful, but there are simulations inside that don't provide added value (should be deleted)                                                                            |
| webots_ros2_epuck            | move            | Maybe move to epuck_ros2                                                                                                                                                    |
| webots_ros2_ur_e_description | use upstream    | Hopefully, the ur_description will be published soon for ROS 2, so we can use the upstream version                                                                          |

PR #269, opened 2021-08-19
Utilize `webots_ros2_driver` in `webots_ros2_mavic` and `webots_ros2_turtlebot`

lukicdarkoo:
The main objective of the PR is to deprecate the `webots_ros2_core` packages in favor of `webots_ros2_driver`. This also involves:
- creating system tests for each converted package,
- deleting unnecessary packages (see #239), and
- changing the `webots_ros2_driver` to rely on Driver instead of Supervisor (I believe we will need to publish additional cpp files to libController to make it work on Windows).

**TODO**
- [x] `webots_ros2_turtlebot`: Add system tests
- [x] `webots_ros2_mavic`: Convert
- [x] `webots_ros2_mavic`: Add system tests


BenjaminHug8 (approved):
The 3 added tests in those changes pass.

PR #278, opened 2021-08-25
fix 273

dzywater:
fix 273

### DIFF
--- a/webots_ros2_driver/src/WebotsNode.cpp
+++ b/webots_ros2_driver/src/WebotsNode.cpp
@@ -55,6 +55,7 @@ namespace webots_ros2_driver
     }
     else
     {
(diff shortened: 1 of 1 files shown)


webots_ros2_driver/src/WebotsNode.cpp
```diff
@@ -55,6 +55,7 @@ namespace webots_ros2_driver
     }
     else
     {
+      mWebotsXMLElement=NULL;
```
lukicdarkoo, line 58:
```suggestion
      mWebotsXMLElement = NULL;
```

lukicdarkoo (approved):
Thank you!

I have just sent you an invitation so you can merge the PR (and run the CI tests)

PR #281, opened 2021-08-27
Fix 280

dzywater:


dzywater:
> > if not above PR, only generate
> > ~\install\bin\webots_ros2_control.dll
> > ~\install\bin\webots_ros2_control_system.dll
> > no
> > ~\install\lib\webots_ros2_control.lib
> > ~\install\lib\webots_ros2_control_system.lib
> 
> Yes, I understand it, but why is it necessary? The plugin works without it.

I want to inherit Ros2ControlSystem/Ros2Control for further development and other ideas.

lukicdarkoo (approved):
I see, thank you. Looking forward to seeing what you are going to make ;)

PR #283, opened 2021-08-30
Print tips: robot_state_publisher started or not

dzywater:
Checkout robot_state_publisher node exists or not. If not, info: joint state publisher is not available.


webots_ros2_driver/src/WebotsNode.cpp
```diff
@@ -112,8 +112,18 @@ namespace webots_ros2_driver
...
+    std::vector<std::string> nodeNames = this->get_node_names();
+    for (int k = 0; k < nodeNames.size(); ++k)
+    {
+      if (nodeNames[k] == "/robot_state_publisher")
+      {
+        setAnotherNodeParameter("robot_state_publisher", "robot_description", mRobot->getUrdf());
+        break;
+      }
+      if (k == nodeNames.size() - 1) 
+        RCLCPP_INFO(get_logger(), "No robot_state_publisher node and joint state publisher is not available.");
```
lukicdarkoo, line 124:
```suggestion
        RCLCPP_INFO(get_logger(), "There is no the `robot_state_publisher` node, the `robot_description` parameter is not set.");
```

webots_ros2_driver/src/WebotsNode.cpp
```diff
@@ -112,8 +112,18 @@ namespace webots_ros2_driver
 
   void WebotsNode::init()
   {
-    setAnotherNodeParameter("robot_state_publisher", "robot_description", mRobot->getUrdf());
-
+    std::vector<std::string> nodeNames = this->get_node_names();
```
lukicdarkoo, line 115:
```suggestion
    const std::vector<std::string> nodeNames = this->get_node_names();
```

lukicdarkoo (changes requested):
Nice, thanks.

I have a few style comments. Also, the `get_node_names` method returns currently available nodes. I am afraid of a situation in which the `robot_state_publisher` starts just after Webots. In general, it should not happen as Webots typically takes more time to load, but in the case of lightweight simulations, the problem may occur.

dzywater:
> Nice, thanks.
> 
> I have a few style comments. Also, the `get_node_names` method returns currently available nodes. I am afraid of a situation in which the `robot_state_publisher` starts just after Webots. In general, it should not happen as Webots typically takes more time to load, but in the case of lightweight simulations, the problem may occur.

The second step is to provide a service which calls setAnotherNodeParameter again. Because someone may start robot_state_publisher after WebotsNode.

lukicdarkoo:
> The second step is to provide a service which calls setAnotherNodeParameter again

That's a good idea!

webots_ros2_driver/src/WebotsNode.cpp
lukicdarkoo: Can this be a const?

lukicdarkoo (approved):
Thank you

PR #284, opened 2021-08-30
Add service set_robot_description

dzywater:


lukicdarkoo:
I was thinking a bit and I believe this should not be included in the `webots_ros2` core as one can e.g.:
```bash
ros2 param set ${NODE} ${PARAMETER} $(ros2 param get /TurtleBot3Burger robot_description)
```

Furthermore, we should consider removing the `setAnotherNodeParameter` method in favor of the approach given above.

dzywater:
> I was thinking a bit and I believe this should not be included in the `webots_ros2` core as one can e.g.:
> 
> ```shell
> ros2 param set ${NODE} ${PARAMETER} $(ros2 param get /TurtleBot3Burger robot_description)
> ```
> 
> Furthermore, we should consider removing the `setAnotherNodeParameter` method in favor of the approach given above. What do you thit is better.

> I was thinking a bit and I believe this should not be included in the `webots_ros2` core as one can e.g.:
> 
> ```shell
> ros2 param set ${NODE} ${PARAMETER} $(ros2 param get /TurtleBot3Burger robot_description)
> ```
> 
> Furthermore, we should consider removing the `setAnotherNodeParameter` method in favor of the approach given above. What do you think?

It is more reasonable. I agree!

dzywater:
> I was thinking a bit and I believe this should not be included in the `webots_ros2` core as one can e.g.:
> 
> ```shell
> ros2 param set ${NODE} ${PARAMETER} $(ros2 param get /TurtleBot3Burger robot_description)
> ```
> 
> Furthermore, we should consider removing the `setAnotherNodeParameter` method in favor of the approach given above. What do you think?

It is more reasonable. I agree!

dzywater:
> I was thinking a bit and I believe this should not be included in the `webots_ros2` core as one can e.g.:
> 
> ```shell
> ros2 param set ${NODE} ${PARAMETER} $(ros2 param get /TurtleBot3Burger robot_description)
> ```
> 
> Furthermore, we should consider removing the `setAnotherNodeParameter` method in favor of the approach given above. What do you think?
I suggest: 
var mWebotsConfig for webots ros2 configuration
var mRobotDescription for robot description

dzywater:
> I was thinking a bit and I believe this should not be included in the `webots_ros2` core as one can e.g.:
> 
> ```shell
> ros2 param set ${NODE} ${PARAMETER} $(ros2 param get /TurtleBot3Burger robot_description)
> ```
> 
> Furthermore, we should consider removing the `setAnotherNodeParameter` method in favor of the approach given above. What do you think?

Besides, I suggest:
var mWebotsConfig for webots ros2 configuration
var mRobotDescription for robot description

dzywater:
> I was thinking a bit and I believe this should not be included in the `webots_ros2` core as one can e.g.:
> 
> ```shell
> ros2 param set ${NODE} ${PARAMETER} $(ros2 param get /TurtleBot3Burger robot_description)
> ```
> 
> Furthermore, we should consider removing the `setAnotherNodeParameter` method in favor of the approach given above. What do you think?

I think setAnotherNodeParameter should be kept and called once when inited. Besides, setAnotherNodeParameter is a general function and can be used to set other types of params.

(1) If robot_state_publisher exists, print tips like:  robot description is set successfully.

(2) If robot_state_publisher not exist, print tips like : robot description is not set and you can set it by "robot_description" param after starting robot_state_publisher.

we should find the reason that robot_state_publisher cannot be returned. Or which function can return it.

PR #285, opened 2021-08-31
Revert `robot_state_publisher` check

lukicdarkoo:
This PR reverts changes made in #283.

The `get_node_names` never returns the `robot_state_publisher` node on Linux. @dzywater we have to find another way to check whether the `robot_state_publisher` node exists.

### DIFF
--- a/webots_ros2_driver/src/WebotsNode.cpp
+++ b/webots_ros2_driver/src/WebotsNode.cpp
@@ -112,18 +112,8 @@ namespace webots_ros2_driver
 
   void WebotsNode::init()
   {
-    const std::vector<std::string> nodeNames = this->get_node_names();
-    for (int k = 0; k < nodeNames.size(); ++k)
-    {
-      if (nodeNames[k] == "/robot_state_publisher")
-      {
-        setAnotherNodeParameter("robot_state_publisher", "robot_description", mRobot->getUrdf());
-        break;
-      }
-      if (k == nodeNames.size() - 1) 
-        RCLCPP_INFO(get_logger(), "There is no the `robot_state_publisher` node, the `robot_description` parameter is not set.");
-    }
-    
+    setAnotherNodeParameter("robot_state_publisher", "robot_description", mRobot->getUrdf());
+
     mStep = mRobot->getBasicTimeStep();
     mTimer = this->create_wall_timer(std::chrono::milliseconds(1), std::bind(&WebotsNode::timerCallback, this));
 


dzywater:
> This PR reverts changes made in #283.
> 
> The `get_node_names` never returns the `robot_state_publisher` node on Linux. @dzywater we have to find another way to check whether the `robot_state_publisher` node exists.

Actually, I also met with this problem on windows10 in one computer on which other ros2 apps also met with some problems(https://github.com/ros2/ros2/issues/1072 && https://github.com/ros2/ros2/issues/1073). But on other computers, it is normal. So I thought it was the environment problem at that time and not studied it further. I will check it on Ubuntu.

dzywater:
@lukicdarkoo 
the way in ros2cli source.
https://github.com/ros2/ros2cli/blob/master/ros2node/ros2node/verb/list.py#L38
https://github.com/ros2/ros2cli/blob/1f4a3aaf7b930830b3fa04773715bf93ea65693e/ros2node/ros2node/api/__init__.py#L59

dzywater:
@lukicdarkoo
the way in ros2cli source.
https://github.com/ros2/ros2cli/blob/master/ros2node/ros2node/verb/list.py#L38
https://github.com/ros2/ros2cli/blob/master/ros2node/ros2node/api/__init__.py#L59

lukicdarkoo:
> @lukicdarkoo
> the way in ros2cli source.
> https://github.com/ros2/ros2cli/blob/master/ros2node/ros2node/verb/list.py#L38
> https://github.com/ros2/ros2cli/blob/master/ros2node/ros2node/api/__init__.py#L59

Instead of checking for the list of available nodes, I would rather handle it in the `setAnotherNodeParameter` method, simply by checking timeout (if there is timeout then the node doesn't exist).

dzywater:
> > @lukicdarkoo
> > the way in ros2cli source.
> > https://github.com/ros2/ros2cli/blob/master/ros2node/ros2node/verb/list.py#L38
> > https://github.com/ros2/ros2cli/blob/master/ros2node/ros2node/api/__init__.py#L59
> 
> Instead of checking for the list of available nodes, I would rather handle it in the `setAnotherNodeParameter` method, simply by checking timeout (if there is timeout then the node doesn't exist).

OK, I agree!

dzywater:
> > @lukicdarkoo
> > the way in ros2cli source.
> > https://github.com/ros2/ros2cli/blob/master/ros2node/ros2node/verb/list.py#L38
> > https://github.com/ros2/ros2cli/blob/master/ros2node/ros2node/api/__init__.py#L59
> 
> Instead of checking for the list of available nodes, I would rather handle it in the `setAnotherNodeParameter` method, simply by checking timeout (if there is timeout then the node doesn't exist).

we cam give up #284 and implement param way in this pr.

BenjaminHug8 (commented):
Version tested and working on the 2 tutorials found [here](https://github.com/cyberbotics/webots_ros2/wiki/Example-TurtleBot3-Burger).

BenjaminHug8 (approved):

PR #286, opened 2021-08-31
Utilize `webots_ros2_driver` in Universal Robot examples

lukicdarkoo:
**Tasks**
- [x] Utilize `webots_ros2_driver`
- [x] Add system tests
- [x] Squash multiple examples
  - `webots_ros2_demos`: Moved to `webots_ros2_universal_robot`
  - `webots_ros2_ur_e_description`: Deleted. URDF obtained from `wb_robot_get_urdf()` is used instead of URDFs provided in the `webots_ros2_ur_e_description` package. Ideally, we will implement [this](https://discourse.ros.org/t/urdf-with-webots/19665) in the future.
  - `webots_ros2_abb`: Deleted. The example wasn't providing any added value.
  - `webots_ros2_universal_robot/universal_robot_multiple.wbt`: This is a very nice-looking simulation. I would like to keep it, but it doesn't provide any value to the Webots ROS 2 integration. Maybe we can place it in `webots-projects`?
- Others:
  - [x] Fix `webots_ros2_control` to set the initial state values to `NaN`
  - [x] Add `streaming` argument to `WebotsLauncher`
  - [x] Record the simulation (with `rosbag2`) only on failure

**Summary**
This PR partially addresses #239

The `webots_ros2_universal_robot` package now provides:
- `robot_launch.py`: An example of `ros2_control` integration.
- `moveit_demo_launch.py`: MoveIt2 integration.
  - Ideally, we should test it, but the Python interface hasn't been ported to ROS 2. I prefer to skip tests for this simulation until the interface is ready (https://github.com/ros-planning/moveit2/issues/314).
- `multirobot_launch.py`: An example of controlling multiple robots (boils down to namespace and environment variables configuration).

**Notes**
- Documentation has to be synced.


BenjaminHug8 (approved):
Works fine on my computer, thanks!
I also adapt the documentation pages related to universal robots.

PR #291, opened 2021-09-01
Make some members protected in WebotsNode.hpp

dzywater:
Make some members protected in WebotsNode.hpp for users' further development and ideas' try
  mRobot 
  mRobotName 
  mRobotDescription 
  mClient s
  setAnotherNodeParameter

### DIFF
--- a/webots_ros2_driver/include/webots_ros2_driver/WebotsNode.hpp
+++ b/webots_ros2_driver/include/webots_ros2_driver/WebotsNode.hpp
@@ -49,12 +49,13 @@ namespace webots_ros2_driver
       std::string mRobotName;
       std::string mRobotDescription;
 
+      rclcpp::Client<rcl_interfaces::srv::SetParameters>::SharedPtr mClient;
+      void setAnotherNodeParameter(std::string anotherNodeName, std::string parameterName, std::string parameterValue);
+
   private:
     void timerCallback();
     std::unordered_map<std::string, std::string> getDeviceRosProperties(const std::string &name) const;
     std::unordered_map<std::string, std::string> getPluginProperties(tinyxml2::XMLElement *pluginElement) const;
-    void setAnotherNodeParameter(std::string anotherNodeName, std::string parameterName, std::string parameterValue);
-    rclcpp::Client<rcl_interfaces::srv::SetParameters>::SharedPtr mClient;
 
     int mStep;
     rclcpp::TimerBase::SharedPtr mTimer;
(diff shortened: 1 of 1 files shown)


webots_ros2_driver/include/webots_ros2_driver/WebotsNode.hpp
```diff
@@ -44,19 +44,21 @@ namespace webots_ros2_driver
     webots::Supervisor *robot() { return mRobot; }
     std::string urdf() const { return mRobotDescription; };
 
+  protected:
+      webots::Supervisor* mRobot;
```
lukicdarkoo, line 48: This is accessible through the `robot()` getter.

webots_ros2_driver/include/webots_ros2_driver/WebotsNode.hpp
```diff
@@ -44,19 +44,21 @@ namespace webots_ros2_driver
...
+      std::string mRobotName;
+      std::string mRobotDescription;
+
```
lukicdarkoo, line 51: This is accessible through the `urdf()` getter.

lukicdarkoo (commented):
Exposing more of API means more maintenance and less flexibility on our side. Could you please tell me your use case?

dzywater:
Oh, sorry, I ignored them suddenly, robot() and urdf() are enough.

lukicdarkoo:
Could you please propose changes to:
https://github.com/cyberbotics/webots_ros2/wiki/Tutorial-Creating-a-Custom-Plugin
so the others are also aware of the methods?

lukicdarkoo:
The parameter name is the same, but the node name is not. To change a parameter, you have to specify a node name as well.

I wouldn't rename it because we eventually want to support full URDF specification:
- https://discourse.ros.org/t/urdf-with-webots/19665
- https://github.com/cyberbotics/urdf2webots/issues/112

dzywater:
> The parameter name is the same, but the node name is not. To change a parameter, you have to specify a node name as well. Therefore, there are no conflicts.
> 
> I wouldn't rename it because we want to implement full URDF support:
> 
>     * https://discourse.ros.org/t/urdf-with-webots/19665
> 
>     * [Question: why not introduce a <webots> tag? urdf2webots#112](https://github.com/cyberbotics/urdf2webots/issues/112)

Wow! That's a good news! 
Which version will plan to implement full URDF support?

lukicdarkoo:
> Which version will plan to implement full URDF support?

We have some ideas on how to implement it, but we haven't settled on anything. Once we do a detailed plan we will be able to estimate the time of release.

dzywater:
> > Which version will plan to implement full URDF support?
> 
> We have some ideas on how to implement it, but we haven't settled on anything. Once we do a detailed plan we will be able to estimate the time of release.

Currently, how could we get webots robot's urdf (Robot::getUrdf) in other ros nodes?

lukicdarkoo:
You read the `robot_description` param. Alternatively, you can make [a simple Python plugin](https://github.com/cyberbotics/webots_ros2/wiki/Tutorial-Creating-a-Custom-Python-Plugin) that publishes a latched topic (or service) with the description.

Issue #287, opened 2021-08-31
Add system tests for TurtleBot SLAM and Navigation tutorials


lukicdarkoo:
Since `Navigate TurtleBot3` and `SLAM with TurtleBot3` tutorials are listed in [the official ROS 2 tutorials page](https://docs.ros.org/en/foxy/Tutorials.html) we should ensure the tutorials are high quality. Therefore, we need a system test to verify whether the Webots-TurtleBot integration works as expected.

PR #292, opened 2021-09-01
Creation of system test for TurtleBot SLAM tutorial

BenjaminHug8:
**Description**
This PR is for creating system test for `TurtleBot` `SLAM` tutorial as the [official tutorial page of ROS2](https://docs.ros.org/en/foxy/Tutorials.html) is pointing to it and `Navigation` tutorial.

This system test is supposed to verify whether the Webots-TurtleBot integration works as expected.

**Related Issues**
This pull-request fixes issue #287 (`Navigation` test will be done in another PR and issue)

**Tasks**
  - [x] System test for SLAM tutorial


lukicdarkoo:
I see that the Rolling CI is failing because the Rolling (and Galactic) TurtleBot3 packages are not released. I have just posted an issue about that:
https://github.com/ROBOTIS-GIT/turtlebot3/issues/778

Meanwhile, a solution might be to "manually" install the `turtlebot3` package for Foxy in the `scripts/ci_after_init.bash` script. Then, in Rolling, we simply ignore the tests.

scripts/ci_after_init.bash
```diff
@@ -1,6 +1,13 @@
...
+ROS_REPO=$2
+
 apt update
 apt install -y wget dialog apt-utils psmisc
 wget https://github.com/cyberbotics/webots/releases/download/R${WEBOTS_VERSION}/webots_${WEBOTS_VERSION}_amd64.deb -O /tmp/webots.deb
 apt install -y /tmp/webots.deb xvfb
+
+if [ "${ROS_DISTRO}" = "foxy" ]; then
+    apt install -y ros-foxy-turtlebot3-cartographer ros-foxy-turtlebot3-navigation2
+fi
```
lukicdarkoo, line 13:
```suggestion
fi

```

webots_ros2_tests/test/test_system_turtlebot_tutorials.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # Webots
+    turtlebot_webots = IncludeLaunchDescription(
+        PythonLaunchDescriptionSource(
+            os.path.join(get_package_share_directory('webots_ros2_turtlebot'), 'launch', 'robot_launch.py')
+        )
+    )
+
+    # Rviz SLAM
+    turtlebot_SLAM = IncludeLaunchDescription(
```
lukicdarkoo, line 48:
```suggestion
    turtlebot_slam = IncludeLaunchDescription(
```
Everything is either upper-case or lower-case

webots_ros2_tests/test/test_system_turtlebot_tutorials.py
```diff
@@ -0,0 +1,85 @@
...
+        rclpy.shutdown()
+
+    def setUp(self):
+        self.__node = rclpy.create_node('driver_tester')
+        self.wait_for_clock(self.__node, messages_to_receive=20)
+
+    def testSLAM(self):
+        from nav_msgs.msg import OccupancyGrid
+
+        def on_map_message_received(message):
```
lukicdarkoo, line 78: To improve the test, it would be useful to check whether there is an obstacle in the grid. The map can be empty and it is not valid.

lukicdarkoo (changes requested):
Thank you, very nice!

There are a few minor issues to fix

scripts/ci_after_init.bash
```diff
@@ -1,6 +1,13 @@
...
 
+ROS_DISTRO=$1
+ROS_REPO=$2
+
 apt update
 apt install -y wget dialog apt-utils psmisc
 wget https://github.com/cyberbotics/webots/releases/download/R${WEBOTS_VERSION}/webots_${WEBOTS_VERSION}_amd64.deb -O /tmp/webots.deb
 apt install -y /tmp/webots.deb xvfb
+
+if [ "${ROS_DISTRO}" = "foxy" ]; then
```
lukicdarkoo, line 11: Here we need a short comment on why this condition exists and why the TurtleBot packages are installed through apt (instead of `package.xml`)

BenjaminHug8: Done in the next commit.

lukicdarkoo:
It's safe to ignore the `testing, rolling` test

scripts/ci_after_init.bash
```diff
@@ -1,6 +1,14 @@
...
 
+ROS_DISTRO=$1
+ROS_REPO=$2
+
 apt update
 apt install -y wget dialog apt-utils psmisc
 wget https://github.com/cyberbotics/webots/releases/download/R${WEBOTS_VERSION}/webots_${WEBOTS_VERSION}_amd64.deb -O /tmp/webots.deb
 apt install -y /tmp/webots.deb xvfb
+
+# The following packages are only yupported by ROS2 Foxy, done here to be sure that ROS_DISTRO is available
```
lukicdarkoo, line 11:
```suggestion
# The following packages are only available in the ROS 2 Foxy distribution. Therefore, we cannot include them in the package.xml, but we have to install them manually here.
```

lukicdarkoo (approved):
Thank you!

lukicdarkoo:
The failing CI is actually a known regression:
- https://discourse.ros.org/t/preparing-for-rolling-sync-2021-09-09/22138
- http://repo.ros2.org/status_page/ros_rolling_default.html?q=REGRESSION

Even the Rolling sync is postponed because of it. We are affected by the regression since the `webots_ros2_turtlebot` package depends on `diff_drive_controller`

Issue #242, opened 2021-07-10
Possible improper call of _wb_device.recognitionDisable()


scottnortman:
Using Webots 2021a with the ROS2 interface, Ubuntu 20.04.

I created numerous cameras in the Webots world and do not need the built in recognition functionality, so I leaver the recognition as NULL.  Images will instead b e streamed via a topic into an alternate recognition node through ROS2.

After launching, I am able to see images as expect in RVIZ2 and other nodes, however I recieve a constant stream of output messages in the launch terminal:

[webots_node] Error: wb_camera_recognition_disable() called on a Camera without Recognition node.

I traced this down to the file 

webots_ros2_core/camera_device.py

If you look at the if /  else statement, with the 'if' on line 147, it checks to see if the camera has recognition, and if not, falls through to the 'else' statement on line 214.  The call inside the 'else' is

self._wb_device.recognitionDisable()

but in my case the recognition is NULL, so there is no instance that can call the method, and hence the error output.  I commented out that line and it resolved the Error output messages.

PR #294, opened 2021-09-03
fix call we_camera_recognition_disable() on NULL recognition node

BenjaminHug8:
**Description**
This PR is to fix a call of wb_camera_recognition_disable() on a Camera without Recognition node in webots_ros2_core/camera_device.py.

**Related Issues**
This pull-request fixes issue #242 

**Affected Packages**
List of affected packages:
  - webots_ros2_core

**Solution**
Fixed by simply checking if the Camera has a Recognition node before call wb_camera_recognition_disable().


lukicdarkoo (approved):
Thank you!

PR #300, opened 2021-09-16
Tesla conversion from Core to Driver

BenjaminHug8:
**Description**
This PR is meant to convert `Tesla` from `core` to `driver`. 

**Additional context**
This PR is linked to the PR [#3668](https://github.com/cyberbotics/webots/pull/3668) of `webots` repository.


webots_ros2_driver/src/PythonPlugin.cpp
```diff
@@ -29,17 +29,31 @@ namespace webots_ros2_driver
...
             R"EOT(
-from webots_ros2_driver_webots.controller import Supervisor
+import sys, os
+import webots_ros2_driver_webots
+
+path = os.path.dirname(webots_ros2_driver_webots.__file__)
+
+sys.path.insert(1, path)
+
+from vehicle import Driver
```
lukicdarkoo, line 39:
```suggestion
sys.path.insert(0, os.path.dirname(webots_ros2_driver_webots.__file__))
from vehicle import Driver
```
Avoid exposing unnecessary variables and group related instructions. Also, why `1` and not `0`?

webots_ros2_driver/src/PythonPlugin.cpp
```diff
@@ -29,17 +29,31 @@ namespace webots_ros2_driver
 
         PyObject *pyWebotsExtraModuleSource = Py_CompileString(
             R"EOT(
-from webots_ros2_driver_webots.controller import Supervisor
+import sys, os
+import webots_ros2_driver_webots
+
```
lukicdarkoo, line 34: We should explain why the path is extended

webots_ros2_driver/src/PythonPlugin.cpp
```diff
@@ -29,17 +29,31 @@ namespace webots_ros2_driver
...
+from vehicle import Driver
 
 class WebotsNode:
     def __init__(self):
-        self.robot = Supervisor.internalGetInstance()
+        self.robot = Driver.internalGetInstanceDriverIfFeasible()
+
+        # Driver.internalGetInstance == Robot.internalGetInstance
+        if self.robot is None :
```
lukicdarkoo, line 46:
```suggestion
        if self.robot is None:
```

webots_ros2_driver/src/PythonPlugin.cpp
```diff
@@ -29,17 +29,31 @@ namespace webots_ros2_driver
...
+            self.robot = Driver.internalGetInstance()
 )EOT",
             "webots_extra", Py_file_input);
+//self.robot = Car.internalGetInstanceDriver()
```
lukicdarkoo, line 50:
```suggestion
//self.robot = Car.internalGetInstanceDriver()
```
Avoid commenting the code

lukicdarkoo (changes requested):
Very good!

However, we should also support Windows and the Windows support should be added in this PR (otherwise the `master` branch will be broken)

webots_ros2_driver/src/Driver.cpp
```diff
@@ -16,12 +16,20 @@
 #include <rclcpp/rclcpp.hpp>
 #include <webots_ros2_driver/WebotsNode.hpp>
 
+#include <string>
+#include <iostream>
+
```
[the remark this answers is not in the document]
BenjaminHug8: No, I forgot to remove them after debugging

webots_ros2_driver/src/PythonPlugin.cpp
BenjaminHug8: Alright for exposing !

I put `1` instead of `0` because I read that the position `0` is risky some times.

Issue #310, opened 2021-10-26
Universal robot system tests fails time to time


BenjaminHug8:
**Describe the Bug**
It seems that a system test of `universal_robot` is failing due to some initialization that fails, thus the test fails due to overtime.

BenjaminHug8:
This has been solved for now with a time sleep in PR #311, need to investigate further.

BenjaminHug8:
Solved in [this PR](https://github.com/cyberbotics/webots_ros2/pull/333).

PR #311, opened 2021-10-27
Fix universal robot system tests random faillure

BenjaminHug8:
**Description**
This PR is to solve the random failure due to time out for the system test of `universal robot`.

**Related Issues**
This pull-request fixes issue #310.


BenjaminHug8:
After adding the sleep time in the controllers and add a bit more of time before time out for the controller manager and the system test, it seems to be solved (check 3 consecutive run).

I also corrected a bit the trajectory for ABB arm.

omichel (approved):
We should investigate why this ugly time.sleep seems to be needed here. There should be a better way to synchronize things as ROS 2 is supposed to deterministic.

PR #322, opened 2021-11-03
Fixed broken LED

omichel:
The LED topic was broken: it was expecting a message of type `ColorRGBA` which is not compatible with all the possibilities of LED devices in Webots (indexed, gradual, RGB). This PR reverts to using a `Int32` message type to be closer to the Webots API and allow all the use cases.

I updated the [tutorial about it](https://github.com/cyberbotics/webots_ros2/wiki/Tutorial-E-puck-for-ROS2-Beginners).


BenjaminHug8 (approved):
Thank you!

BenjaminHug8 (approved):
Thank you!

PR #328, opened 2021-11-05
Fix buildfarm withour ROS_DISTRO defined

BenjaminHug8:
**Description**
This should solved the issue on the buildfarm as `ROS_DISTRO` is not defined.
Including the `ros_environnement` package should solve this according to this [PR](https://github.com/ros/rosdistro/issues/29465).


omichel (approved):
That seems very good.

PR #345, opened 2021-12-06
Feature URDF import on the fly

BenjaminHug8:
**Description**
This PR aims to ease the use of URDF files in ROS 2 - Webots simulation.

In a launch file we can start Webots and use the `WebotsLauncher` classe from `webots_ros2_driver` to add a Supervisor robot that will spawn URDF robots.

A second launcher will launch all the ros2 nodes and will communicate with the Supervisor `spawner` to give the information about the URDF robot.

`webots_ros2_universal_robot` is supposed to be an example how to use this feature.

**Affected Packages**
List of affected packages:
  - webots_ros2_driver
  - webots_ros2_universal_robot

**Tasks**
  - [x] Add the supervisor `spawner`.
  - [x] Create the spawner ROS node.
  - [x] Create a universal launcher to start a simulation with the supervisor.
  - [x] Adapt the `webots_ros2_universal_robot` Ur5e robot to show an example how to use this.
  - (CANCELED) Make the importer able to use directly Xacro files in addition to URDF files.
  - [x] Convert the 3 launchers for `webots_ros2_universal_robot` showing how to use the URDF Ur5e robot (see below).
  - [x] Launcher 1: single Ur5e robot.
  - [x] Launcher 2: single Ur5e robot with MoveIt2.
  - [x] Launcher 3: one Ur5e robot with one ABB robot.
  - [x] Deal with relative paths included in the world file, as it will be copied to add the `spawner`.

**Before merging and in order**
  - [x] Finish [this PR on urdf2webots](https://github.com/cyberbotics/urdf2webots/pull/146) as it proposes a new feature used by the `spawner`.
  - [x] Create another PR based on this one to take care of #368, finish it and merge it into this one.

**After this PR has been merged**
- [ ] Show in a tutorial how one can start the Webots launcher and the other ROS 2 nodes launcher separately. The 2nd launcher can be stopped and relaunch at will in order to refresh the URDF robots in the simulation. (Can talk quickly about Xacro, like [here](https://github.com/cyberbotics/webots_ros2/pull/345#issuecomment-1014646288) ).

**Additional context**
Related to this [PR](https://github.com/cyberbotics/urdf2webots/pull/142) of `urdf2webots`.


webots_ros2_urdf_converter/package.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0"?>
```
lukicdarkoo, line 1: I believe it is better to use the on-the-fly conversion in the `webots_ros2_universal_robot` package because we already need URDF there for MoveIt2. See:
https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_universal_robot/resource/moveit_ur5e_description.urdf

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
```diff
@@ -82,3 +124,57 @@ def __init__(self, output='screen', world=None, gui=True, mode='realtime', strea
...
+
+    def execute(self, context: LaunchContext):
+        # Check if the user wants to convert URDF files into robots
+        if self.__robots:
+            worldPath = self.__world.perform(context)
+            if not worldPath:
+                sys.exit('World file not specified (has to be specified with world=path/to/my/world.wbt')
+
+            worldCopy = worldPath[:-4] + '_with_URDF_robot.wbt'
+            shutil.copyfile(worldPath, worldCopy)
```
lukicdarkoo, line 136: I am afraid this will not work as a package might be installed in the`/opt` directory for which we need `sudo` privileges. Instead, we should put it in the `/temp` directory (there is a Python API for getting a path to a temporary directory).

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
```diff
@@ -82,3 +124,57 @@ def __init__(self, output='screen', world=None, gui=True, mode='realtime', strea
...
+                robot_name = robot.get('name')
+                robot_translation = robot.get('translation')
+                robot_rotation = robot.get('rotation')
+
+                if not file_input:
+                    sys.exit('URDF file not specified (has to be specified with \'urdf_location\': \'path/to/my/robotUrdf.urdf\'')
+                if not robot_name:
+                    sys.stderr.write('Robot name not specified (should be specified if more than one robot is present with \'name\': \'robotName\'\n')
+
+                convert2urdf(inFile=file_input, worldFile=worldCopy, robotName=robot_name, initTranslation=robot_translation, initRotation=robot_rotation)
```
lukicdarkoo, line 151: The added `worldFile` argument here breaks the "abstraction". We should use the urdf2webots tools as is (with small changes). Instead, I would change the `convert2urdf` function to return a URDF file as a string, or we can use it as is but write the output to a temp file.

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
```diff
@@ -82,3 +124,57 @@ def __init__(self, output='screen', world=None, gui=True, mode='realtime', strea
...
+
+        return super().execute(context)
+
+    def getDriversList(self):
```
lukicdarkoo, line 155: We can still spawn controllers as we were doing it so far. Why do we need this?

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
```diff
@@ -82,3 +124,57 @@ def __init__(self, output='screen', world=None, gui=True, mode='realtime', strea
             ],
             **kwargs
         )
+
+    def execute(self, context: LaunchContext):
+        # Check if the user wants to convert URDF files into robots
+        if self.__robots:
+            worldPath = self.__world.perform(context)
```
lukicdarkoo, line 131:
```suggestion
            world_path = self.__world.perform(context)
```
Underscore

lukicdarkoo (dismissed):

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
Timple: Is it possible to spawn a robot into a world? Instead of 'hardcoding' it in the worldfiles?

lukicdarkoo: We can make it possible. Why do you think it is an advantage?

webots_ros2_urdf_converter/package.xml
BenjaminHug8: Good idea!
I believed there were only Webots configuration URDF file in this example as I had not noticed the `moveit_ur5e_description.urdf` file.

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
Timple: In my opinion robots operate in a world, instead of being part of it. But that might just be semantics.

But a typical flow for us is to have a repo with worlds (our office). And several of our robots can operate in this world.
Now we need N worlds for the N robot types we have. (Or script our way around that).

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
BenjaminHug8: As the conversion and the `driver` node need the URDF file I wanted to ask the user to give it only once, but this required to specify more information in the `robotsList`.

Do you think it is better to stick as it is now? So there would be less differences between launching a `driver` node for an URDF / non-URDF robot ?

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
lukicdarkoo: No, you don't need N worlds for the N robot types, one world will be enough. It will generate new worlds in the background, no need to worry about that. The only difference is that you don't need a spawner.

From user's perspective it will be:
```py
webots = WebotsLauncher(
    world=PathJoinSubstitution([package_dir, 'worlds', world]),
    robots=[{'urdf': '/path/to/urdf.urdf', 'translation': '0 0 0'}]
)
```

instead of:
```py
webots = WebotsLauncher(
    world=PathJoinSubstitution([package_dir, 'worlds', world])
)
```

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
BenjaminHug8: So you mean that `convert2urdf` will still convert URDF to PROTO but also either append a robot to a world file or return the robot in VRML as a string?

lukicdarkoo: `convert2urdf` should never append to a world file, it should be handled outside of the `webots2urdf` package.

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
Timple: Ah true, I had the previous situation in mind. So this effort already helps!

Than the typical use-case is covered.

Only thing left on my double agenda: we have a 'game' going on between colleagues where we can race each other in webots. It would be nice if anyone could 'join' a world without having to restart the world for all players.

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
lukicdarkoo: I see. However, I don't think it is a big problem for a user to pass URDF two times. It is more consistent with the current behavior.

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
lukicdarkoo: > Only thing left on my double agenda: we have a 'game' going on between colleagues where we can race each other in webots. It would be nice if anyone could 'join' a world without having to restart the world for all players.

I would love to see the game :)

You can still create a supervisor controller which will spawn robots on demand. It is more effort though. We assumed it is a very uncommon use-case, so figured out it is not very important to support it out-of-the-box. Can you think of another use-case in which it is a common practice to spawn a robot dynamically?

Timple: Is quickly iterating on your robot a use-case?

You can leave the (rather heavy) world running in webots and continue developing on the robot. 
Continuously restarting webots takes time unfortunately.

BenjaminHug8:
The URDF importer feature should be ready now (at least it works fine locally), but as stated in the description I have to finish 2 other PRs in order for this one to be finished.

Timple:
I can't wait for this feature! So I went ahead and tried it.

meshes relative to the original world file cannot be located anymore. Probably because of the temporary world created in another location?

BenjaminHug8:
> I can't wait for this feature! So I went ahead and tried it.

Thank you for your interest :)

> meshes relative to the original world file cannot be located anymore. Probably because of the temporary world created in another location?

Yes you are right, I forgot this scenario, thanks!

Timple:
> drop the Xacro support as it will be not user friendly to use and user can easily handle this himself.

Agreeing with this, but perhaps this can be added as an example somewhere?
```python
robot_description = xacro.process_file('/path/to/somebot.urdf.xacro', mappings={"input": "new_input"}).toxml()
spawn_urdf_somebot = URDFSpawner(urdf_path=robot_description)
```

BenjaminHug8:
@omichel May I ask you to review already once this PR? (`foxy-main` failure is the RViz random crash in CI, I have created an issue about it).

After that I will move the `/clock` topic to `Ros2Supervisor` as planned.

webots_ros2_driver/include/webots_ros2_driver/WebotsNode.hpp
```diff
@@ -50,10 +54,13 @@ namespace webots_ros2_driver
     void setAnotherNodeParameter(std::string anotherNodeName, std::string parameterName, std::string parameterValue);
     rclcpp::Client<rcl_interfaces::srv::SetParameters>::SharedPtr mClient;
 
-    std::string mRobotName;
     std::string mRobotDescription;
     bool mSetRobotStatePublisher;
 
+    bool waitForURDFRobotRemoved = false;
```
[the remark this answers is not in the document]
omichel: Also the naming here is not very clear... Do you mean that we should wait for the the removal of a URDF robot? Or we are waiting for it?

webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
```diff
@@ -0,0 +1,119 @@
...
+        robot = request.robot
+
+        robot_name = robot.name if robot.name else ''
+
+        if robot_name == '':
+            self.get_logger().info('Ros2Supervisor cannot import an unnamed URDF robot. Please specifiy it with name="" in the URDFSpawner object.')
+            response.success = False
+            return response
+        if robot_name in self.__urdf_robots_list:
+            self.get_logger().info('The URDF robot name "' + str(robot_name) + '" is already used ny another robot! Please specifiy another unique name.')
```
omichel, line 63:
```suggestion
            self.get_logger().info('The URDF robot name "' + str(robot_name) + '" is already used by another robot! Please specify a unique name.')
```

webots_ros2_driver/webots_ros2_driver/urdf_spawner.py
```diff
@@ -0,0 +1,63 @@
...
+    return
+
+class URDFSpawner(ExecuteProcess):
+    def __init__(self, output='screen', name=None, urdf_path=None, translation='0 0 0', rotation='0 0 1 0', normal=False, box_collision=False, init_pos=None, **kwargs):
+        command = [
+                'ros2',
+                'service',
+                'call',
+                '/spawn_urdf_robot',
+                'webots_ros2_msgs/srv/SetWbURDFRobot',
```
omichel, line 36: This naming looks odd... Can't we move the Wb part?
```suggestion
                'webots_ros2_msgs/srv/SetWbURDFRobot',
```

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
```diff
@@ -82,3 +95,70 @@ def __init__(self, output='screen', world=None, gui=True, mode='realtime', strea
             ],
             **kwargs
         )
+
+    def execute(self, context: LaunchContext):
+        # User can give a PathJoinSubstitution world ...
+        if isinstance(self.__world, PathJoinSubstitution):
+            world_path = self.__world.perform(context)
+            context.launch_configurations['world'] = self.__world_copy.name
+        # or an absolute path world
```
omichel, line 104: I don't understand this comment? Is it on the right line? What does it mean?

webots_ros2_universal_robot/README.md
```diff
@@ -4,3 +4,127 @@ This package provides an interface between ROS2 and the [UR3e, UR5e and UR10e si
...
+<joint name="${prefix}tool0_palm" type="fixed">
+    <parent link="${prefix}tool0"/>
+    <child link="${prefix}palm"/>
+    <origin xyz="0 0.045 0" rpy="1.5708 0 0"/>
+</joint>
+```
+- Update accordingly to your package all the paths.
+
+#### The [robotiq-3f-gripper_finger_articulated_macro.xacro](https://github.com/ros-industrial/robotiq/blob/kinetic-devel/robotiq_3f_gripper_visualization/cfg/robotiq-3f-gripper_finger_articulated_macro.xacro) file:
+- Update accordingly to your package all the paths.
```
omichel, line 130:
```suggestion
- Update all the paths accordingly to your package.
```

webots_ros2_universal_robot/launch/multirobot_launch/world_multirobot_launch.py
```diff
@@ -0,0 +1,56 @@
...
+    webots = WebotsLauncher(
+        world=PathJoinSubstitution([package_dir, 'worlds', world])
+    )
+
+    ros2_supervisor = Ros2SupervisorLauncher()
+
+    return LaunchDescription([
+        DeclareLaunchArgument(
+            'world',
+            default_value='armed_robots.wbt',
```
omichel, line 44: What about this name?
```suggestion
            default_value='robotic_arms.wbt',
```

webots_ros2_driver/include/webots_ros2_driver/WebotsNode.hpp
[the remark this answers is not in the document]
BenjaminHug8: We are waiting for `Ros2Supervisor` to remove the Robot node string from the world.

webots_ros2_driver/src/Driver.cpp
```diff
@@ -17,10 +17,15 @@
 #include <webots_ros2_driver/WebotsNode.hpp>
 #include <webots/vehicle/Driver.hpp>
 
-int main(int argc, char **argv)
+bool SIGINTReceived = false;
```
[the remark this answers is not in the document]
BenjaminHug8: I will change that, let me know what you think about it in the next review.

webots_ros2_driver/webots_ros2_driver/urdf_spawner.py
BenjaminHug8: That is the way to indicate the type of the srv used. [See an example here](https://docs.ros.org/en/foxy/Tutorials/Topics/Understanding-ROS2-Topics.html#ros2-topic-pub)

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
BenjaminHug8: It goes with line 100, but I will put these 2 comments on 1 line to make it clearer.

webots_ros2_driver/webots_ros2_driver/urdf_spawner.py
omichel: Yes, but I meant the name of this srv is odd to me.

webots_ros2_driver/include/webots_ros2_driver/WebotsNode.hpp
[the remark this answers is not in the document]
omichel: In that case, `mWaitingForUrdfRobotToBeRemoved` sounds clearer to me.

webots_ros2_driver/webots_ros2_driver/urdf_spawner.py
BenjaminHug8: Ah ok I understand now, as discussed together I will rename the msgs to remove the `Wb` (the name of the package `webots_ros2_msgs` is sufficient to show the Webots specificity of these messages ).

BenjaminHug8:
Hum `foxy-main` keeps failing due to the RViz crash, I will investigate that...

BenjaminHug8:
@lukicdarkoo I re ask you a review as you requested changes that will refrain me to merge this PR, but you are not forced to review this PR :)

omichel (changes requested):
As discussed, it would be nice to move the signal handling code inside the WebotsNode class.

webots_ros2_driver/src/WebotsNode.cpp
```diff
@@ -39,6 +39,8 @@ namespace webots_ros2_driver
   const char *gPluginInterface = "webots_ros2_driver::PluginInterface";
   const char *gPluginInterfaceName = "webots_ros2_driver";
 
+  bool WebotsNode::mShutdownSignalReceived = false;
```
omichel, line 42: This shouldn't be exposed:
```suggestion
  bool gShutdownSignalReceived = false;
  
  void handleSigint()
  {
    bool gShutdownSignalReceived = true;
  }
```

webots_ros2_driver/include/webots_ros2_driver/WebotsNode.hpp
```diff
@@ -42,6 +42,8 @@ namespace webots_ros2_driver
     void init();
     webots::Supervisor *robot() { return mRobot; }
     std::string urdf() const { return mRobotDescription; };
+    static void customSignalHandler(int sig);
+    static bool mShutdownSignalReceived;
```
omichel, line 46:
```suggestion
    static void handleSignals();
```

webots_ros2_driver/src/Driver.cpp
```diff
@@ -29,6 +27,17 @@ int main(int argc, char **argv)
   else
     robot = new webots::Supervisor();
 
+  // Replace the signal handler for the WebotsNode and the robot by a custom one
+  //signal(SIGINT, webots_ros2_driver::WebotsNode::customSignalHandler);
+  signal(SIGINT, webots_ros2_driver::WebotsNode::customSignalHandler);
```
omichel, line 32:
```suggestion
  // Let WebotsNode handle the system signals
  webots_ros2_driver::WebotsNode::handleSignals();
```

webots_ros2_driver/src/WebotsNode.cpp
```diff
@@ -240,4 +244,9 @@ namespace webots_ros2_driver
     request->parameters.push_back(parameter);
     mClient->async_send_request(request);
   }
+
+  void WebotsNode::customSignalHandler(int sig)
+  {
+    mShutdownSignalReceived = true;
+  }
```
omichel, line 251:
```suggestion
  void WebotsNode::handleSignals()
  {
    signal(SIGINT, handleSigint);
  }
```

webots_ros2_driver/src/WebotsNode.cpp
```diff
@@ -216,16 +219,17 @@ namespace webots_ros2_driver
 
   void WebotsNode::timerCallback()
   {
+    if (mShutdownSignalReceived && !mWaitingForUrdfRobotToBeRemoved){
```
omichel, line 222:
```suggestion
    if (gShutdownSignalReceived && !mWaitingForUrdfRobotToBeRemoved) {
```

omichel (changes requested):
I believe the management of the signal should be fully integrated inside the WebotsNode class, see my suggestions.

BenjaminHug8:
> I believe the management of the signal should be fully integrated inside the WebotsNode class, see my suggestions.

I had to change it a bit. With this implementation `Driver` does not transmit the signal to `WebotsNode`.

webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
```diff
@@ -0,0 +1,125 @@
...
+import webots_ros2_driver_webots
+from rclpy.time import Time
+from rclpy.node import Node
+from rclpy.qos import qos_profile_services_default
+from rosgraph_msgs.msg import Clock
+from std_msgs.msg import String
+from urdf2webots.importer import convert2urdf
+from webots_ros2_msgs.srv import SpawnUrdfRobot
+
+# As Ros2Supervisor need the controller library, we extend the path here
```
omichel, line 34:
```suggestion
# As Ros2Supervisor needs the controller library, we extend the path here
```

webots_ros2_driver/webots_ros2_driver/urdf_spawner.py
```diff
@@ -0,0 +1,66 @@
...
+    def __init__(self, output='log', name=None, robot_description=None, translation='0 0 0', rotation='0 0 1 0', normal=False, box_collision=False, init_pos=None, **kwargs):
+        message = '{robot: {'
+
+        if name:
+            message += 'name: "' + name + '",'
+        if robot_description:
+            # Prepare the robot_description to be send via command.
+            robot_description = robot_description.replace("'", "\'")
+            robot_description = robot_description.replace('"', '\\"')
+            message += 'robot_description: "\\\n' + robot_description + '",'
```
BenjaminHug8, line 39: I am not a huge fan of this, but I have not found another way to do it without a ROS node

BenjaminHug8:
It seems that the `rolling` packages are not available now.

webots_ros2_driver/webots_ros2_driver/urdf_spawner.py
BenjaminHug8: Moreover, I think we should also give the option to use the path of the URDF file, in case the URDF file uses relative paths (it will be impossible for the converter tool to guess the correct location).

omichel: Yes, giving the path would certainly be useful...

webots_ros2_driver/webots_ros2_driver/urdf_spawner.py
```diff
@@ -0,0 +1,66 @@
...
+class URDFSpawner(ExecuteProcess):
+    def __init__(self, output='log', name=None, robot_description=None, translation='0 0 0', rotation='0 0 1 0', normal=False, box_collision=False, init_pos=None, **kwargs):
+        message = '{robot: {'
+
+        if name:
+            message += 'name: "' + name + '",'
+        if robot_description:
+            # Prepare the robot_description to be send via command.
+            robot_description = robot_description.replace("'", "\'")
+            robot_description = robot_description.replace('"', '\\"')
```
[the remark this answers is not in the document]
BenjaminHug8: I will double check but as my string is using `"` quotes as delimiters I think I have to do it like this, otherwise the `"` contained in the robot description won't be different than the `"` at the end of the string.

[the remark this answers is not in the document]
omichel: OK, thank you for checking. My proposed change still remains a valid simplification.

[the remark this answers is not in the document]
BenjaminHug8: Thanks for the reminder, I was focus on the question of the `\\"` !

webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def __remove_urdf_robot_callback(self, message):
+        robotName = message.data
+
+        if robotName in self.__urdf_robots_list:
+            robot_node = None
+
+            for id_node in range(self.__insertion_robot_place.getCount()):
+                node = self.__insertion_robot_place.getMFNode(id_node)
+                node_name_field = node.getField("name")
```
omichel, line 107: We usually prefer single quotes:
```suggestion
                node_name_field = node.getField('name')
```

webots_ros2_driver/webots_ros2_driver/urdf_spawner.py
```diff
@@ -0,0 +1,70 @@
...
+# limitations under the License.
+
+"""This process simply sends urdf information to the Spawner through a service."""
+
+from launch.actions import ExecuteProcess
+
+
+def get_webots_driver_node(event, driver_node):
+    """Return the driver node in case the service response is successful."""
+    if "success=True" in event.text.decode().strip():
```
omichel, line 24:
```suggestion
    if 'success=True' in event.text.decode().strip():
```

webots_ros2_driver/webots_ros2_driver/urdf_spawner.py
```diff
@@ -0,0 +1,70 @@
...
+        return driver_node
+    print("WARNING: the Ros2Supervisor was not able to spawn this URDF robot.")
```
omichel, line 26:
```suggestion
    print('WARNING: the Ros2Supervisor was not able to spawn this URDF robot.')
```

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
```diff
@@ -48,14 +57,19 @@ def __init__(self, output='screen', world=None, gui=True, mode='realtime', strea
         webots_path = os.path.join(webots_path, 'webots')
 
         mode = mode if isinstance(mode, Substitution) else TextSubstitution(text=mode)
-        world = world if isinstance(world, Substitution) else TextSubstitution(text=world)
+
+        self.__world_copy = tempfile.NamedTemporaryFile(mode="w+", suffix='_world_with_URDF_robot.wbt', delete=False)
```
omichel, line 61:
```suggestion
        self.__world_copy = tempfile.NamedTemporaryFile(mode='w+', suffix='_world_with_URDF_robot.wbt', delete=False)
```

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
```diff
@@ -77,3 +91,69 @@ def __init__(self, output='screen', world=None, gui=True, mode='realtime', strea
...
+
+        # Update relative paths in the world
+        with open(self.__world_copy.name, 'r') as file:
+            content = file.read()
+
+        for match in re.finditer('url\s*\[?\s*\"(.*?)\"', content):
+            url_path = match.group(1)
+
+            # Absolute path or Webots relative path or Web paths
+            if os.path.isabs(url_path) or "webots://" in url_path or "http://" in url_path or "https://" in url_path:
```
omichel, line 113:
```suggestion
            if os.path.isabs(url_path) or 'webots://' in url_path or 'http://' in url_path or 'https://' in url_path:
```

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
```diff
@@ -77,3 +91,69 @@ def __init__(self, output='screen', world=None, gui=True, mode='realtime', strea
...
+
+    def _shutdown_process(self, context, *, send_sigint):
+        # Remove copy of the world and the corresponding ".wbproj" file
+        if self.__world_copy:
+            self.__world_copy.close()
+            if os.path.isfile(self.__world_copy.name):
+                os.unlink(self.__world_copy.name)
+
+            path, file = os.path.split(self.__world_copy.name)
+            world_copy_secondary_file = os.path.join(path, "." + file[:-1] + "proj")
```
omichel, line 143:
```suggestion
            world_copy_secondary_file = os.path.join(path, '.' + file[:-1] + 'proj')
```

omichel (changes requested):

omichel (approved):
That looks good to me.
Thank you.

PR #346, opened 2021-12-10
Fix controller_manager update rate

### LINKED ISSUE #343: TurtleBot3 simulation doesn't work at 50Hz as specified
Given:
- https://github.com/cyberbotics/webots_ros2/blob/c1b3a055f9b3d931066083e561c3dbee5b386c3a/webots_ros2_turtlebot/resource/ros2control.yml#L3
- https://github.com/cyberbotics/webots_ros2/blob/c1b3a055f9b3d931066083e561c3dbee5b386c3a/webots_ros2_turtlebot/resource/ros2control.yml#L19

We should receive Odometry messages at 50Hz, but instead, we get them at 30Hz.

This is reasonable as a simulation with a `basicTimeStep` at 32ms cannot deliver that high frequency. However, reducing the `basicTimeStep` increases the frequency to 60Hz which is wrong.

This indicates there is a bug in `webots_ros2_control`.

lukicdarkoo:
Fixes #343

There is not much we can do to tackle this. Basically:
- If basicTimeStep doesn't match the update rate then we notify the user about that.
- If a user uses low basicTimeStep then we skip some updates.

**Tasks**
- [x] Fix the update_rate
- [x] Fix copyright 
- [x] Fix CI


lukicdarkoo:
- Rolling CI fail: Rolling never executes the `generate_test_description()` function. It seems something is updated, but I cannot find what. Investigating...
- Galactic (main): This should be temporary as Galactic (testing) works fine.

lukicdarkoo:
We got affected by https://github.com/ros2/launch_ros/pull/278. I believe we didn't use the tests in the correct way.

lukicdarkoo:
- `CI / industrial_ci (main, galactic) (pull_request)`: It seems they pushed a wrong header (looks fixed in `testing`)
- `CI / industrial_ci (testing, foxy) (pull_request)`: Standard random error caused by RViz and the Turtlebot3 package.

webots_ros2_turtlebot/worlds/turtlebot3_burger_example.wbt
```diff
@@ -4,6 +4,7 @@ WorldInfo {
     "Simple TurtleBot3 Burger simulation that can be controlled by a ROS."
   ]
   title "ROS2 simulation of the TurtleBot3 Burger robot"
+  basicTimeStep 20
```
BenjaminHug8, line 7: Same as `Tiago`.

webots_ros2_turtlebot/resource/ros2control.yml
```diff
@@ -16,5 +16,4 @@ diffdrive_controller:
     wheel_separation: 0.160
     wheel_radius: 0.033
 
-    publish_rate: 50.0
```
BenjaminHug8, line 19: Same as other `ros2_control.yml`.

webots_ros2_tiago/worlds/default.wbt
```diff
@@ -3,7 +3,7 @@ WorldInfo {
   info [
     "An office break room, surrounded by office desks."
   ]
-  basicTimeStep 16
+  basicTimeStep 20
```
BenjaminHug8, line 6: Same as `rats_life`.

webots_ros2_epuck/worlds/rats_life_benchmark.wbt
```diff
@@ -3,7 +3,7 @@ WorldInfo {
   info [
     "Rat's life pattern"
   ]
-  basicTimeStep 16
+  basicTimeStep 20
```
BenjaminHug8, line 6: Same as `e-puck`.

webots_ros2_epuck/resource/ros2_control.yml
```diff
@@ -16,5 +16,4 @@ diffdrive_controller:
     wheel_separation: 0.05685
     wheel_radius: 0.02
 
-    publish_rate: 50.0
```
BenjaminHug8, line 19: Do we simply remove it as it has no effect and I will put it again when my PR on `ros2control` is accepted, or should we keep it anyway?

webots_ros2_tiago/resource/ros2_control.yml
```diff
@@ -16,7 +16,6 @@ diffdrive_controller:
     wheel_separation: 0.5
     wheel_radius: 0.12
 
-    publish_rate: 50.0
```
BenjaminHug8, line 19: Same as other `ros2_control.yml`.

webots_ros2_epuck/worlds/epuck_world.wbt
```diff
@@ -4,6 +4,7 @@ WorldInfo {
     "Simple e-puck simulation that can be controlled with ROS2."
   ]
   title "ROS2 simulation of the e-puck robot"
+  basicTimeStep 20
```
BenjaminHug8, line 7: Do we have to keep this `basicTimeStep`? I think it was set only for testing?

BenjaminHug8 (changes requested):
Seems all good, I just have 1 question and 1 suggestion that occur multiple times.

webots_ros2_epuck/resource/ros2_control.yml
lukicdarkoo: I would rather put it later.

webots_ros2_epuck/worlds/epuck_world.wbt
lukicdarkoo: If we want to have 50Hz publish rate then we have to put it.

webots_ros2_epuck/resource/ros2_control.yml
BenjaminHug8: Fine for me then.

webots_ros2_epuck/worlds/epuck_world.wbt
BenjaminHug8: Ok fine for me too.

BenjaminHug8 (approved):
Thank you.

PR #351, opened 2021-12-16
Move `rclcpp_lifecycle` dependency to `webots_ros2_control`

lukicdarkoo:
`webots_ros2_driver` doesn't depend on `rclcpp_lifecycle`, but `webots_ros2_control` only


BenjaminHug8 (approved):
Thank you.

PR #353, opened 2021-12-16
Make Navigation2 optional

lukicdarkoo:
Make sure Tiago simulation doesn't crash on `get_package_share_directory('nav2_bringup')`. Note that Navigation2 is not released to Rolling, so we cannot explicitly make it a dependency.

### DIFF
--- a/webots_ros2_tiago/launch/robot_launch.py
+++ b/webots_ros2_tiago/launch/robot_launch.py
@@ -115,9 +115,7 @@ def generate_launch_description():
                 ('map', nav2_map),
                 ('use_sim_time', use_sim_time),
             ],
-            condition=launch.conditions.IfCondition(use_nav)
-        )
-        )
+            condition=launch.conditions.IfCondition(use_nav)))
 
     slam_toolbox = Node(
         parameters=[{'use_sim_time': use_sim_time}],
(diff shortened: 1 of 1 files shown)


BenjaminHug8 (dismissed):
Thank you.

webots_ros2_tiago/launch/robot_launch.py
```diff
@@ -106,14 +107,17 @@ def generate_launch_description():
...
+    if 'nav2_bringup' in get_packages_with_prefixes():
+        optional_nodes.append(IncludeLaunchDescription(
+            PythonLaunchDescriptionSource(os.path.join(
+                get_package_share_directory('nav2_bringup'), 'launch', 'bringup_launch.py')),
+            launch_arguments=[
+                ('map', nav2_map),
+                ('use_sim_time', use_sim_time),
+            ],
+            condition=launch.conditions.IfCondition(use_nav)
+        )
```
omichel, line 119: Indentation looks wrong here:
```suggestion
            )
```

webots_ros2_tiago/launch/robot_launch.py
```diff
@@ -106,14 +107,17 @@ def generate_launch_description():
...
+        )
```
omichel, line 120: This kind of indentation looks better to me:
```suggestion
            condition=launch.conditions.IfCondition(use_nav)))
```

lukicdarkoo: VSCode did it in a weird way :)

omichel (approved):
Thank you.

PR #354, opened 2021-12-17
Feature add GPS speed vector

BenjaminHug8:
**Description**
To be consistent with `webots_ros` we need to add a topic for the `speed_vector`.

**Details**
Currently the norm of the velocity is published on `gps/velocity` but the name `velocity` is supposed to refer to a vector and `speed` is supposed to refer to the norm of the `velocity` vector.

Thus I changed the publishing of the velocity norm on the `gps/speed` topic and the velocity on the `gps/velocity` topic.
I also added a test for this topic.

This branch targets this other [PR](https://github.com/cyberbotics/webots_ros2/pull/316) as the other PR is not yet merged into `develop` and this feature relies on the `R2022a` lib-controller.


BenjaminHug8:
- `foxy/testing` has a random fail due to RViz.
- `galactic/main` fails as we made a patch for the ros2control update present in `galactic/testing` but not yet released in `galactic/main`.

omichel (changes requested):
The distinction between speed and velocity is not clear and not consistent with the Webots API, we should use speed_vector instead of velocity.

omichel (approved):
Perfect. Thank you.

PR #355, opened 2021-12-17
Merge develop into enu conversion

BenjaminHug8:
**conflicts**

- `webots_driver/webots`


lukicdarkoo:
Why the intermediate branch? We could simply merge develop into #316 directly? We need intermediate branches only when syncing protected branches

BenjaminHug8:
> Why the intermediate branch? We could simply merge develop into #316 directly? We need intermediate branches only when syncing protected branches

I wanted to deal with possible conflicts in this intermediate branch but I will do directly the merge directly into #316 then.

PR #357, opened 2021-12-20
Changelogs update version 1.2.0

BenjaminHug8:
**Description**
In order to prepare the `1.2.0` release I updated the `CHANGELOG` files.


lukicdarkoo (changes requested):
It looks good, but we should also change the following line before release:
https://github.com/cyberbotics/webots_ros2/blob/54005fe5fbf0ae4d9bb12479076ee9ba3687f664/scripts/ci_after_init.bash#L8

We can do it in this PR

lukicdarkoo (approved):
Thank you!

PR #360, opened 2021-12-21
Simplify C++ plugins

lukicdarkoo:
Export libController include files, so plugin developers don't have to include them explicitly


BenjaminHug8 (approved):
Thank you.

PR #361, opened 2021-12-21
Allow access to a supervisor instance

### LINKED ISSUE #349: Extended ros2_driver interface with Python plugin does not access Supervisor API
**Describe the Bug**
We de not have access to the `Supervisor API` from a Python plugin for `ros2_driver` node (see tesla or mavic for example).

This issue comes from [here](https://github.com/cyberbotics/webots_ros2/blob/c1b3a055f9b3d931066083e561c3dbee5b386c3a/webots_ros2_driver/src/PythonPlugin.cpp#L46) where we get a `Robot` instance instead of a `Supervisor` one when the robot is not a car.

We should change the function used to return a `Supervisor` instance.

lukicdarkoo:
Fixes #349

---

The CI may fail because of the MoveIt dependency, but you can check the previous CI run


webots_ros2_turtlebot/worlds/turtlebot3_burger_example.wbt
```diff
@@ -21,6 +21,7 @@ TurtleBot3Burger {
   controllerArgs [
     ""
   ]
+  supervisor TRUE
```
BenjaminHug8, line 24: May I ask why we should change that?

lukicdarkoo: I missed deleting it :/ Thanks

BenjaminHug8 (approved):
Not so sure why there is a diff in `turtlebot3_burger_example.wbt` but it seems there is no change in fact. I let you just double check and otherwise it is all good thanks!

omichel:
> Not so sure why there is a diff in `turtlebot3_burger_example.wbt` but it seems there is no change in fact. I let you just double check and otherwise it is all good thanks!

It could be a mixture of Linux and Windows carriage returns. It is worth checking it.

lukicdarkoo:
It is the carriage return:
```
$ diff turtlebot3_burger_example.wbt turtlebot3_burger_example2.wbt --strip-trailing-cr
$ diff turtlebot3_burger_example.wbt turtlebot3_burger_example2.wbt 
1,18c1,18
< #VRML_SIM R2022a utf8
< WorldInfo {
<   info [
<     "Simple TurtleBot3 Burger simulation that can be controlled by a ROS."
<   ]
<   title "ROS2 simulation of the TurtleBot3 Burger robot"
<   basicTimeStep 20
< }
< Viewpoint {
<   orientation -0.2878951570967451 -0.16506369377382343 0.9433293992651676 4.177335774513749
<   position 9.141595121588473 4.8923306199482335 3.8476210861955575
<   follow "TurtleBot3Burger:LDS-01"
< }
< TexturedBackground {
```

However, I am not sure what caused it, I simply saved the world.

omichel:
> However, I am not sure what caused it, I simply saved the world.

In Atom, you can set the line ending (preferably to LF instead of CRLF) before you save so that it is consistent.

lukicdarkoo:
> > However, I am not sure what caused it, I simply saved the world.
> 
> In Atom, you can set the line ending (preferably to LF instead of CRLF) before you save so that it is consistent.

I don't recall whether I save in editor the first time, but I have just tried saving it in Webots and the file hasn't changed (except the nodes I changed in the scene tree)

omichel:
@lukicdarkoo: The end line encoding is wrong for `turtlebot3_burget_examples.wbt`. It is `CRLF` (Windows) whereas it should be `LF` (Linux). Can you please fix it?

lukicdarkoo:
Oh, sorry, everything should be LF now.

PR #362, opened 2021-12-22
Update urdf2webots

BenjaminHug8:


BenjaminHug8:
`Foxy/testing` fails due to random fail of RViz.

PR #367, opened 2022-01-05
Fix LiDAR transform

### LINKED ISSUE #366: Wrong transform of LiDAR sensor
**Describe the Bug**
LiDAR scan data outputs not correct. I think the reason is that the coordinate system was changed to ENU/FLU in the R2022a version.

**Steps to Reproduce**
1. Install webots R2022a (binary)
2. Source build webots_ros2(branch: master)
3. Follow the turtlebot3 navigation2 instruction described [here](https://github.com/cyberbotics/webots_ros2/wiki/Navigate-TurtleBot3)
4. Check LaserScan data on rviz2

**Expected behavior**
LiDAR scan data should be matched with map.

**Affected Packages**
List of affected packages:
  - webots_ros2_turtlebot3
  - webots_ros2_driver
  - ros-foxy-turtlebot3-navigation2

**Screenshots**
When initialpose is set up, Lidar scan visualised as follow:
image

**System**
 - Webots Version: R2022a
 - ROS Version: Foxy Fossa(debian install)
 - Operating System: Linux Ubuntu 20.04

lukicdarkoo:
Fixes #366

**Tasks**
- [x] Fix LiDAR transform
- [x] Add more LiDAR tests
- [x] Verify the related tutorials manually
- [x] Verify point cloud messages
- [x] Revert temporary deleted MoveIt dependency
- Not related problems:
  - [x] Fix Webots version
  - [x] Fix the missing transforms

**Note**
- The tests are passing without the MoveIt dependency, so it seems there is a temporary problem on the ROS side.
- It might be possible that we inverted the order of the LiDAR scan in Webots.
- We should do another release with this fix as soon as possible (before ROS syncs).

### DIFF
--- a/webots_ros2_core/webots_ros2_core/utils.py
+++ b/webots_ros2_core/webots_ros2_core/utils.py
@@ -33,8 +33,8 @@ from launch.action import Action
 from launch.launch_context import LaunchContext
 
 
-MINIMUM_VERSION_STR = 'R2021c'
-TARGET_VERSION_STR = 'R2021c'
+MINIMUM_VERSION_STR = 'R2022a'
+TARGET_VERSION_STR = 'R2022a'
 
 
 @functools.total_ordering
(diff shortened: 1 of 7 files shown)


omichel (approved):
Thank you.
Can you check whether we inverted the order of the lidar scans in Webots?
I agree once it is merged, we should prepare a new release.

lukicdarkoo:
> Can you check whether we inverted the order of the lidar scans in Webots?

I have just checked and Webots returns measurements from left to right (as before and somehow natural). In ROS, it is expected to send values counterclockwise (mathematically more accurate):
http://docs.ros.org/en/noetic/api/sensor_msgs/html/msg/LaserScan.html (check a description of the header field)

PR #376, opened 2022-01-19
Prepare version 1.2.2

BenjaminHug8:
Prepare new releases to fix the `MoveIt` dependency.


BenjaminHug8:
`foxy-testing` failure is due to Rviz that crashed.

omichel (approved):

Issue #375, opened 2022-01-19
"WebotsNode" is designed to set "robot_description" parameter of "robot_state_publisher" node which is not always useful.


TaoYibo1866:
https://github.com/cyberbotics/webots_ros2/blob/99c090403b29e9102ad8f5afc475f6e99d84ae17/webots_ros2_driver/src/WebotsNode.cpp#L115
https://github.com/cyberbotics/webots_ros2/blob/99c090403b29e9102ad8f5afc475f6e99d84ae17/webots_ros2_driver/src/WebotsNode.cpp#L229-L240

It's very kind that webots_ros2_driver try to set the description file once a robot_state_publisher is launched. However, if i wrote my own urdf and want robot_state_publisher to load that, webots_ros2_driver may override it and set the description file to "mRobot->getUrdf()" which is generated from vrml.

The reason i want to load a urdf without webots dependency is that i want my project to be able to switch between simulation and real world. I have an rviz launch file which runs robot_state_publisher, loads my own urdf and starts rviz2. This works in real world scenario. However, in simulation, when i launch webots and the forementioned launch file together, webots_ros_driver override my urdf parameter and cause rviz2 not showing the right model.

I am not sure i understand the codes correctly. Perhaps this is not the cause of my problem. Hope there is document to help us better understanding the behaviour of webots_ros2_driver, thanks.

BenjaminHug8:
Hi @TaoYibo1866, there is probably a way to get back information from

https://github.com/cyberbotics/webots_ros2/blob/99c090403b29e9102ad8f5afc475f6e99d84ae17/webots_ros2_driver/src/WebotsNode.cpp#L44

So we can use it in the call of the function `setAnotherNodeParameter()` you mentioned.

That is setup like this in the `webots_ros2_universal_robot` `robot_launch` launch file. 

https://github.com/cyberbotics/webots_ros2/blob/99c090403b29e9102ad8f5afc475f6e99d84ae17/webots_ros2_universal_robot/launch/robot_launch.py#L38
https://github.com/cyberbotics/webots_ros2/blob/99c090403b29e9102ad8f5afc475f6e99d84ae17/webots_ros2_universal_robot/launch/robot_launch.py#L64-L73

Does it sound ok for your case?
If yes I will need you to check that it works in your case when I will have implement this.

TaoYibo1866:
[...]

PR #378, opened 2022-01-21
Create an option for "webots_ros2_driver" to publish "robot_description" for "robot_state_publisher" node

BenjaminHug8:
**Description**
Currently `webots_ros2_driver` automatically publish a `robot_description` obtained with `mRobot->getUrdf()` (so the URDF obtained is corresponding either to the PROTO used or to the robot node in the world file) for the `robot_state_publisher` node.

The issue is that `webots_ros2_driver` override the parameter `robot_description`, so a user cannot set it himself. It would be preferable to have an option to specify if `webots_ros2_driver` should publish this `robot_description` in case someone is using a PROTO and in the case the user has an URDF file, he can specify `robot_description` for the `robot_state_publisher` node in the launch file.

**Related Issues**
This pull-request fixes issue #375

### DIFF
--- a/webots_ros2_driver/include/webots_ros2_driver/WebotsNode.hpp
+++ b/webots_ros2_driver/include/webots_ros2_driver/WebotsNode.hpp
@@ -52,6 +52,7 @@ namespace webots_ros2_driver
 
     std::string mRobotName;
     std::string mRobotDescription;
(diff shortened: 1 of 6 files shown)


BenjaminHug8:
`foxy-testing` failure is due to random Rviz crash.

PR #384, opened 2022-01-25
Merge mast into develop

BenjaminHug8:
No conflicts

### DIFF
--- a/scripts/ci_before_init_embed.bash
+++ b/scripts/ci_before_init_embed.bash
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-export WEBOTS_RELEASE_VERSION=2022a
+export WEBOTS_RELEASE_VERSION=2022b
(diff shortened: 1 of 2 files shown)


scripts/ci_before_init_embed.bash
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-export WEBOTS_VERSION=2022a
+export WEBOTS_RELEASE_VERSION=2022a
```
omichel, line 3: Shouldn't it be 2022b here?
```suggestion
export WEBOTS_RELEASE_VERSION=2022b
```

BenjaminHug8: Yes I wanted to do it in another PR but it is better to do it in this one. Thanks.

omichel (approved):

omichel (approved):

PR #390, opened 2022-01-31
Fix Rviz for CI tests

BenjaminHug8:
**Description**
Rviz in CI may crash as it use the Qt libraries from Webots instead of the ones already installed.


omichel (approved):
Great.

PR #396, opened 2022-02-04
Fix Rviz crash with qt in CI

BenjaminHug8:
**Description**
Attempt to solve the random crashes of Rviz in CI:
```
[rviz2-78] loaded library "/usr/lib/x86_64-linux-gnu/qt5/plugins/platforms/libqxcb.so"
[rviz2-78] qt.qpa.xcb: could not connect to display 
[rviz2-78] qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
[rviz2-78] This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
[rviz2-78] 
[rviz2-78] Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, xcb.
```


BenjaminHug8:
As discussed with @omichel we will discard the test using RViz in the CI while searching for a solution.

omichel (approved):

Issue #397, opened 2022-02-04
Re-enable the CI tests for turtlebot once RViz does not randomly crash anymore


BenjaminHug8:
**Describe the Bug**
For now we have disable the CI tests for turtlebot scenarios as RViz randomly.

Waiting on https://github.com/ros-industrial/industrial_ci/issues/772 to know if we can re-enable those tests.

PR #400, opened 2022-02-07
Re enable the RViz tests in the CI

BenjaminHug8:
**Description**
Attempt to re enable the Rviz tests in the CI.

**Related Issues**
This pull-request fixes issue #397


BenjaminHug8:
For anyone wanting to pursue this PR, the issue is that RViz relies on Qt as Webots.

It seems that there is a race condition, my guess is that it depends on the order when RViz and Webots are launched.

ygoumaz:
Closed in favor of #638.

Issue #424, opened 2022-03-29
Tiago example:Slam can't work


discleave:
**Describe the Bug**
'TIAGo SLAM' in 'TIAGo Example' can't build the map

**Steps to Reproduce**
1. execute 'ros2 launch webots_ros2_tiago robot_launch.py rviz:=true slam:=true'
2. See error 'No map recieived' in rviz
3. execute 'ros2 run teleop_twist_keyboard teleop_twist_keyboard'
4. See error. I can control TIAGo by keyboard, but the map can't be built

**Expected behavior**
I could control TIAGo to build a map.

**Screens
![2022-03-29 09-05-28屏幕截图](https://user-images.githubusercontent.com/49744208/160512000-f080b7dc-b56b-4adb-a443-7949eeaf3771.png)
![error2](https://user-images.githubusercontent.com/49744208/160510401-a60f5069-fd7b-42a9-bb1b-3ca85880bb78.png)

**System**
 - Webots Version: 2022a
 - ROS Version:foxy
 - Operating System: Ubuntu 20.04.3

ad-daniel:
Personally it appears to be working for me, the map topic is published for me, but I do have the LaserRangeScan warning you mention. Did you follow the install procedure described here? https://github.com/cyberbotics/webots_ros2/wiki/Build-and-Install

ygoumaz:
I am closing this issue. Feel free to reopen if needed.

PR #425, opened 2022-03-29
Fix laser publisher

ad-daniel:
**Description**
Addresses: https://github.com/cyberbotics/webots_ros2/issues/424

Currently on the tiago example, slam_toolbox complains about the number of range readings not matching.
The change was introduced here https://github.com/cyberbotics/webots_ros2/pull/367, however I'm not certain why. As far as I can tell, for slam_toolbox the range should be inclusive, so `[-0.5*FOV, 0.5*FOV]`, which is also how it was implemented in [webots_ros2_core ](https://github.com/cyberbotics/webots_ros2/blob/2ad205c9118c91ca442ac052073bb81333561c8a/webots_ros2_core/webots_ros2_core/devices/lidar_device.py#L131-L133). @lukicdarkoo do you recall perhaps why the change was done?


lukicdarkoo:
I believe I started from #212, so I didn't notice the implementation is different from the one in `webots_ros2_core`. The solution looks good to me.

omichel (approved):
Thank you.

PR #426, opened 2022-04-01
Allow custom ros2_control hardware_interface

lukicdarkoo:
The `webots_ros2_control::Ros2ControlSystem` hardware_interface is generic and it should work for most users. However, sometimes users need to implement a custom `hardware_interface` (e.g. to achieve a certain control behavior). This PR exports the necessary headers for custom interfaces (if I get time I may write a tutorial about it).

**Affected Packages**
  - webots_ros2_control


ad-daniel (approved):
Thank you! and yes, a tutorial would certainly be appreciated

PR #415, opened 2022-03-09
Fix position + velocity control in ros2_control

lukicdarkoo:
**Description**
This PR solves two bugs (sorry for putting everything in a single PR):
- Allows combining position and velocity control in the ros2_control interface. This is often used with robotic arms, e.g. without it moveit_servo prints singularity warnings all over the place.
- Corrects the focal length in the camera_info topic. The focal length is now calculated in the same way as in the range finder plugin.

**Affected Packages**
  - webots_ros2_driver

PR #428, opened 2022-04-07
Fix motor without sensor

lukicdarkoo:
I believe I introduce the bug in #415


ad-daniel (approved):
Thank you, didn't even notice it.

PS: is there a reason we are using a different coding style compared to Webots? Should we enforce the same clang in ROS2 as I did for ROS?

lukicdarkoo:
> PS: is there a reason for ROS2 we are using a different coding style compared to Webots? Should we enforce the same clang in ROS2 as I did for ROS?

It would be a good idea :) There is no strict C++ coding imposed right now

PR #436, opened 2022-04-25
bump python version to 3.10 for rolling

ijnek:
**Description**
Use Python 3.10 for Rolling on Ubuntu Jammy to fix current CI failure.
**This requires the auto-generated python310 directory in https://github.com/cyberbotics/webots-libcontroller/tree/R2022b/lib/linux-gnu before being merged**, (related to https://github.com/cyberbotics/webots/pull/4474)

Current error log when compiling on Ubuntu Jammy:
```
--- stderr: webots_ros2_driver
  CMake Error at /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
    Could NOT find PythonLibs (missing: PYTHON_LIBRARIES PYTHON_INCLUDE_DIRS)
    (Required is exact version "3.8")
  Call Stack (most recent call first):
    /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
    /usr/share/cmake-3.22/Modules/FindPythonLibs.cmake:310 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
    CMakeLists.txt:20 (find_package)
  
  
  ---
```

**Related Issues**
No related issue

**Affected Packages**
List of affected packages:
  - webots_ros2_driver

PR #438, opened 2022-04-27
Change call to spawner.py executable to spawner

ijnek:
**Description**
``spawner.py`` was deprecated in ROS2 Galactic and removed in ROS2 Rolling in https://github.com/ros-controls/ros2_control/commit/ec0ec193cc09e449d5525e34e38040432fce5511.

It should be replaced by ``spawner``.
Galactic supports both ``spawner.py`` and ``spawner``, so we only have to deal with Foxy. (https://github.com/ros-controls/ros2_control/blob/67bf0a59a6908d9bfd385a05fdcfe3944c7bbb07/controller_manager/CMakeLists.txt#L58)

**Related Issues**
No related issue

**Affected Packages**
List of affected packages:
  - webots_ros2_epuck
  - webots_ros2_tiago
  - webots_ros2_turtle
  - webots_ros2_universal_robot

**Additional context**
This error is showing up in [industrial_ci (main, rolling)](https://github.com/cyberbotics/webots_ros2/runs/6185995564?check_suite_focus=true) of https://github.com/cyberbotics/webots_ros2/pull/436

### DIFF
--- a/webots_ros2_epuck/launch/robot_launch.py
+++ b/webots_ros2_epuck/launch/robot_launch.py
@@ -42,9 +42,11 @@ def generate_launch_description():
     controller_manager_timeout = ['--controller-manager-timeout', '50']
     controller_manager_prefix = 'python.exe' if os.name == 'nt' else ''
 
+    use_deprecated_spawner_py = 'ROS_DISTRO' in os.environ and os.environ['ROS_DISTRO'] == 'foxy'
+
     diffdrive_controller_spawner = Node(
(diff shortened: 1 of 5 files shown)


omichel (approved):
It looks good to me.
Thanks.

omichel:
Unfortunately, the CI tests on rolling still fail...

ijnek:
Yep, I think we're getting closer though! I had a look into those failures, but couldn't quite figure out why those tests were failing.

PR #444, opened 2022-05-24
Synchronize libcontroller

ad-daniel:
**Description**
Due to the recent change on the webots side, develop was broken. It seems that merging master into develop brings up additional issues which make the controllers die immediately, so I've removed it from this PR and will investigate why separately


omichel (approved):
Thank you.

PR #459, opened 2022-07-20
Fix master nightly

ad-daniel:
**Description**

Due to the merge of webots develop into master, nightlies for R2022a are no longer created/available so we need to target R2022a itself.


omichel (approved):
Thank you.

PR #487, opened 2022-10-19
Automatic libcontroller sync

ad-daniel:
**Description**

Proof of concept for automatic creation of PRs when a sync of libcontroller is necessary.
It's also possible to have a workflow (in libcontroller repo) that triggers when something gets merged such that it dispatches an event to this repo, triggering a workflow here which opens the PR. For our purpose however that would be overkill as the libcontroller repo gets updated at most once per day (when the nightlies are created), if even that. It's sufficient to check daily.

### DIFF
--- a/.github/workflows/libcontroller_updater.yml
+++ b/.github/workflows/libcontroller_updater.yml
@@ -1,15 +1,20 @@
 name: Libcontroller Submodule Updater
 
 on:
-  schedule:
-    - cron:  '0 09 * * *'
-
-env:
-  SUBMODULE_TARGET_BRANCH: R2022b
-  WEBOTS_ROS2_BRANCH: master
+  #schedule:
+  #  - cron:  '0 09 * * *'
+  pull_request:
+    types: [labeled, unlabeled]
 
 jobs:
   automatic-submodule-sync:
+    strategy:
+    matrix:
+      include:
+        - submodule_target_branch: R2022b
+          webots_ros2_branch: master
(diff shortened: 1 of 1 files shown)


ad-daniel:
The process seems to have worked when manually triggering it (https://github.com/cyberbotics/webots_ros2/pull/488), will check if the schedule run works too when merged.

.github/workflows/libcontroller_updater.yml
```diff
@@ -0,0 +1,55 @@
...
+on:
+  schedule:
+    - cron:  '0 10 * * *'
+
+jobs:
+  automatic-submodule-sync:
+    strategy:
+      matrix:
+        include:
+          - submodule_target_branch: R2022b
```
ygoumaz, line 12: Shouldn't the target branch for master be R2022b-revision-1 on the libcontroller side ?

ygoumaz (changes requested):
Very nice ! This seems to work perfectly. 

**Side note**: We may also create tags on the libcontroller repo for versions instead of branches. This way there would be a master and a develop branch there too, and versions could be accessed using tags. But we will perform this task when we change the binaries to source files instead.

.github/workflows/libcontroller_updater.yml
ad-daniel: yes indeed

ygoumaz (approved):
Perfect, thank you!

PR #475, opened 2022-09-26
fix issue #462 ros2_supervisor executable

PhilippPolterauer:
**Description**
fixes the Issue #462 

**Related Issues**
This pull-request fixes issue [ros2_supervisor crashing!](https://github.com/cyberbotics/webots_ros2/issues/462#top)
#462

**Affected Packages**
List of affected packages:
  - webots_ros2_driver

**Tasks**
Add the list of tasks of this PR.
  - [x] make ros2_supervisor.py executable

PR #490, opened 2022-10-20
Port WSL changes to master branch

ad-daniel:
**Description**

Ditto.

I can't test it, but contains the diff between develop and master.


ygoumaz:
Nice I will test it asap!

ad-daniel:
> `ros2_supervisor.py` is updated, even though the diff is empty. I suspect the line endings to change. Have you tried to run a few examples in Linux from this branch?

It's a change of permissions introduced here: https://github.com/cyberbotics/webots_ros2/pull/475
Either way yes, it shouldn't show up, I've reverted it. No, I didn't test on linux

ygoumaz:
I tested in the WSL environment and it works fine. I only encounter a problem with `ros_supervisor.py`. The line endings are in the same format as other python files and haven't been changed any PR but for a weird reason I need to update them to CRLF format.

I get the same problem in all branches, so it may be a problem with my configuration.

ygoumaz (approved):

PR #493, opened 2022-10-20
Adjust nightly timings and concurrency

ad-daniel:
**Description**

The libcontroller sync appears to have worked as intended.

This PR:
- Change the time when nighty tests are run (currently the time set is fairly overlapping with the time when the deb itself is generated, so might lead to odd runs if there are delays/slow machines)
- Change the time the sync check is done (early morning, so we have the full day to do something about it if something breaks)
- Make so that workflows currently being run are canceled if a new push occurs (currently it keeps running and a new run is launched for every commit)


ygoumaz (approved):
Thank you!

PR #502, opened 2022-11-01
Enhancement: allow world reset on every samples

ygoumaz:
Currently only the universal robot example can be reset from Webots. Once reset the controllers are able to respawn and connect to each robot again. This PR updates all examples so that their controllers can respawn and reconnect after a reset.


omichel (approved):
Thank you.

ygoumaz:
Tests failures are not related to this PR. I will merge it anyway.

PR #506, opened 2022-11-02
Prepare for version `2022.1.3` update

ygoumaz:
Ditto.


ygoumaz:
The `Changelog.rst` files are a bit messy. Some changes are only given in the webots_ros2 package, while others are duplicated in other packages files. We should decide how to make this more homogeneous and logical.

I suggest to put changes related to each sub-packages in their own `Changelog.rst` files and then list all of them in the `webots_ros2` one to have a global view of what has been changed in the package.

omichel (approved):
Thank you.

PR #508, opened 2022-11-07
Add Python API compatibility and libController compilation

ygoumaz:
Follow-up to #503. Fresh start and rebased to develop.

This PR has 2 goals:
- [x]  Add compatibility with the new Webots Python API introduced in R2023a:
  - [x] Get rid of the C++ controller API in the package. Rely only on the C and Python API.
  - [x] Copy binaries to `lib/controller` folder instead of `lib/`
  - [x] Set `WEBOTS_HOME` to the `webots_ros2` installation folder (instead of Webots) to load the correct libController.
- [x] The current [libController repository](https://github.com/cyberbotics/webots-libcontroller) storing the binaries of the libController is replaced with committed source files. This way the source files can be compiled with the package and ensure compatibility with ARM architecture.


ygoumaz:
I don't think it is possible to test the buildfarm, without making a proper release and PR on ROS Distro. So we should first merge this PR and then make a release to see if the branch solution works.

ad-daniel (commented):
Looks good to me. 
When trying to update the submodule it seems to fail 

```
Fetched in submodule path 'webots_ros2_driver/webots', but it did not contain 717732e9a1bf21b559b8f834fa5cae94d9f116ee. Direct fetching of that commit failed.
```

did you delete the branch by any chance?

Also, to test it in the build farm it will be necessary to merge develop into master and create a release from there. Might wanna make sure develop is up to date with master beforehand

ygoumaz:
If your submodule is still pointing to the libcontroller repository, it may require a few more operations to change the folder to the correct tree and find the commit. The branch and commit exist and are correct, see here: https://github.com/cyberbotics/webots/commit/717732e9a1bf21b559b8f834fa5cae94d9f116ee

ygoumaz:
And yes we should merge #516 and #523 to master first. Then merge master in develop. Finally merge this PR and merge develop in master. 

In addition, we annonced that the webots_ros2_core package would be removed with R2023a. We should probably also create a new PR to delete it before merging develop to master (or directly target master after all the above operations).

ad-daniel:
what about the dummy branch? Shouldn't this PR target develop directly or develop needs other changes?

ad-daniel:
> These branches will be protected and synchronized automatically with master and develop branch of Webots when the libController gets an update.

wouldn't that create countless merge conflicts?

ygoumaz:
Yes, if we try to synchronize "classically", it would be impossible to resolve conflicts. But we can probably push directly to the branch, as it is done on the libcontroller repository right now, what do you think?

ygoumaz:
I have replaced the submodule by commited files, as discussed.

ad-daniel (approved):
I've only checked the files outside driver, I'll take your word the webots folder is complete, it compiles without errors for me

PR #510, opened 2022-11-08
Fix camera focal length

lukicdarkoo:
It seems that FoV in Webots is actually horizontal FoV. With these values, I get results quite close to the ground truth.

It would be good to double-check this.

### DIFF
--- a/webots_ros2/CHANGELOG.rst
+++ b/webots_ros2/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog for package webots_ros2
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+2022.1.4 (2022-XX-XX)
+------------------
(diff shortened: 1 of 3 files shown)


ad-daniel (approved):
According to the docs the camera indeed returns the horizontal field of view https://cyberbotics.com/doc/reference/camera#field-summary, and the focal length is indeed computed as you did for [ROS1](https://github.com/cyberbotics/webots/blob/251035cbd9ee2730095382f119c4782d9e52eb47/projects/default/controllers/ros/RosCamera.cpp#L127), thank you

ygoumaz:
@lukicdarkoo can you please add a line in the `CHANGELOG.rst` files of `webots_ros2_driver` and `webots_ros2` about your modification?

ad-daniel (approved):
Thank you

ygoumaz (approved):
Nice! Thank you.

PR #512, opened 2022-11-10
Fix compilation

ad-daniel:
**Description**

Fixes compilation and syncs libcontroller (needs https://github.com/cyberbotics/webots/pull/5454 to be pushed to libcontroller first)

### DIFF
--- a/webots_ros2_driver/CMakeLists.txt
+++ b/webots_ros2_driver/CMakeLists.txt
@@ -103,14 +103,6 @@ else()
   )
 endif()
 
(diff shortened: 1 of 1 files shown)


ygoumaz:
FYI to make the Python API compatible with webots_ros2, you will have to ament_python_install both packages (python/vehicle) and (python/controller). Also you will have to update the Ros2Supervisor.py file to import both modules.

ad-daniel:
see https://github.com/cyberbotics/webots/pull/5454

PR #527, opened 2022-11-21
Allow custom motor-encoder pair

lukicdarkoo:
**Description**
Currently, the encoder is automatically associated. However, in some cases, we want to use an encoder from another wheel (passive wheels). This PR allows users to choose an encoder in URDF as e.g.:
```xml
<joint name="left_motor">
    <state_interface name="position" />
    <command_interface name="velocity" />
    <param name="sensor">left_encoder</param>
</joint>
```

**Affected Packages**
  - webots_ros2_driver
  - webots_ros2_control

In addition, this PR makes `ros2_supervisor.py` executable. This is necessary, I am not sure how it used to work.


ygoumaz (changes requested):
Thank you! Could you also please add an entry in [webots_ros2/CHANGELOG.rst](https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2/CHANGELOG.rst), which contains all the changes made to the package?

Concerning the executable supervisor, it was already done in #475. It was probably reverted by mistake in another PR.

lukicdarkoo:
I am not sure why tests are failing between the first and second run

ad-daniel:
We are in the process of making a release on the webots side, so the nightlies the workflow uses are not being created currently

ygoumaz:
I have updated your branch with master, so the tests should pass now. However, because now the API used in `webots_ros2` is C and not C++ anymore, I needed to fix some conflicts. Please take a look at the changes to be sure I didn't break anything in your new code.

lukicdarkoo:
> Please take a look at the changes to be sure I didn't break anything in your new code.

It still works, thank you

PR #535, opened 2022-11-24
Upgrade to R2023a (develop)

ygoumaz:
ditto.


ad-daniel (changes requested):
Thank you. It's probably the good time to update the copyrights too. For the webots files it will probably come with the automatic sync, but I noticed we still have stuff mentioning 2021

ad-daniel (approved):
Thank you

PR #523, opened 2022-11-17
Update calculation of `CameraRecognitionObject`

AKreutz:
Calculation now conforms to the new camera coordinate frame introduced with Webots R2022b.

Signed-off-by: Andreas Kreutz <andreas.kreutz@hotmail.de>

**Description**
With the release of Webots R2022b, the coordinate frame for cameras has been changed. This means that the calculation of `CameraRecognitionObject` messages needs to be updated. 

**Related Issues**
This pull-request fixes issue #517 

**Affected Packages**
List of affected packages:
  - webots_ros2_driver
  - webots_ros2_tests

PR #570, opened 2022-12-22
Resolving Deprecation Warning with Header File Replacement

ygoumaz:
Replaces deprecated `<tf2_geometry_msgs/tf2_geometry_msgs.h>` with `<tf2_geometry_msgs/tf2_geometry_msgs.hpp>` in Rolling and Humble. This change was introduced in #523 and was causing a deprecation warning when compiling the package on Humble and Rolling.


ad-daniel (approved):
Thank you

PR #573, opened 2022-12-23
WSL error: local PROTOs cannot be loaded

ygoumaz:
On WSL, worlds referring to local protos could not be loaded because the path is not correctly modified in the world file.

### DIFF
--- a/webots_ros2_driver/webots_ros2_driver/webots_launcher.py
+++ b/webots_ros2_driver/webots_ros2_driver/webots_launcher.py
@@ -157,7 +157,10 @@ class WebotsLauncher(ExecuteProcess):
             if os.path.isabs(url_path) or url_path.startswith('webots://') or url_path.startswith('http://') or url_path.startswith('https://'):
                 continue
 
-            new_url_path = '"' + os.path.split(world_path)[0] + '/' + url_path + '"'
+            new_url_path = os.path.split(world_path)[0] + '/' + url_path
+            if self.__is_wsl:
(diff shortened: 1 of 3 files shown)


webots_ros2_driver/webots_ros2_driver/webots_launcher.py
```diff
@@ -157,7 +157,10 @@ def execute(self, context: LaunchContext):
             if os.path.isabs(url_path) or url_path.startswith('webots://') or url_path.startswith('http://') or url_path.startswith('https://'):
                 continue
 
-            new_url_path = '"' + os.path.split(world_path)[0] + '/' + url_path + '"'
+            new_url_path = os.path.split(world_path)[0] + '/' + url_path
```
ad-daniel, line 160: If the relative urls in the world file (captured into `url_path`) have backward slashes instead of forward ones, does the subsequent command still work?

ad-daniel (changes requested):
Just a doubt

ad-daniel (approved):

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
ygoumaz: As discussed, backward slashes are not used by Webots in world file. They are allowed if used in pairs (`..\\protos\\Example.proto`) but I don't think it worth converting all backslashes to forward ones before executing `wslpath`.

PR #574, opened 2022-12-23
Fix rats life example launcher

ygoumaz:
The name of the epuck in the rats life example is not consistent with the one in the basic epuck example. The robot_launch.py script is failing for the rats life example


ad-daniel (approved):
Thank you

PR #578, opened 2022-12-28
Replace macOS Docker installation by UTM Virtual Machine

ygoumaz:
The new macOS solution on Docker doesn't allow the support for RViz2. 

The macOS solution is now going to be based on a UTM Virtual Machine. Webots will still be started on the macOS host to benefit from hardware acceleration. The difference with Docker is that the VM comes with its own screen and can display RViz2 (without hardware acceleration).

This PR replaces the previous TCP host address given by Docker by the one from the VM host machine, so that the TCP client running in the VM can connect to the host local simulation server.

The new installation procedure is located here: [macOS-Installation-Guide-VM](https://github.com/cyberbotics/webots_ros2/wiki/macOS-Installation-Guide-VM)

This PR also fixes a bug when passing arguments to the TCP client, preventing certain examples to launch (e.g. `webots_ros2_tiago`).

### DIFF
--- a/webots_ros2_driver/scripts/webots_tcp_client.py
+++ b/webots_ros2_driver/scripts/webots_tcp_client.py
@@ -18,10 +18,24 @@
 
 import os
 import socket
+import subprocess
 import sys
 import time
 
-HOST = 'host.docker.internal'  # Connect to host of the container
+
+def get_host_ip():
+    try:
+        output = subprocess.run(['ip', 'route'], check=True, stdout=subprocess.PIPE, universal_newlines=True)
+        for line in output.stdout.split('\n'):
+            fields = line.split()
+            if fields and fields[0] == 'default':
+                return fields[2]
+        sys.exit('Unable to get host IP address.')
+    except subprocess.CalledProcessError:
+        sys.exit('Unable to get host IP address. \'ip route\' could not be executed.')
+
+
+HOST = get_host_ip()  # Connect to host of the container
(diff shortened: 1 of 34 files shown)


webots_ros2_driver/scripts/webots_tcp_client.py
```diff
@@ -18,15 +18,35 @@
...
+
+HOST = get_host_ip()  # Connect to host of the VM
 PORT = 2000  # Port to connect to
 
 tcp_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-launch_arguments = sys.argv[1]
-world_name = sys.argv[2]
+launch_arguments = ''
+for arg in sys.argv:
+    if arg.endswith('.py') or arg == '':
```
omichel, line 44: It should never happen that `arg == ''`, isn't it?
```suggestion
    if arg.endswith('.py'):
```

ygoumaz: It can be empty, because they are passed as `conditionalSubstitutions` here: https://github.com/cyberbotics/webots_ros2/blob/896517292cd230739c087dbb9e071545fcb2fb5f/webots_ros2_driver/webots_ros2_driver/webots_launcher.py#L106-L110

which means that if it is not required, it will be passed as an empty string. For example, if no stream is required, `--stream` is empty.

omichel (approved):
Thank you.

PR #584, opened 2023-01-03
Fix naming of ros2_control hardware type

ad-daniel:
**Description**

The member variable has been renamed


ad-daniel:
The fix is needed, but we should wait for a rolling sync before merging it. Doesn't make much sense to add further conditionals to distinguish MAIN and TESTING

ygoumaz:
Nice fix! I was also thinking about waiting for the Rolling sync to fix the issue. If the problem gets to main in a short time, we will need to re-create the package again. Is there a reason not to wait?

nuclearsandwich:
If at all possible, I would recommend against using the `ROS_REPO` variable, which I think is particular to indusrial_ci and not part of the `ros_environment` package, in order to conditionalize build configurations. I think it would be better to detect directly the version of the dependency which changed its interface. Using indirect information will always be prone to misconfigurations when the information you're keying on (such as which repositories are active) changes. For example the upcoming sync to Rolling will synchronize the contents of the testing repo to main, which also means that this change will be propagated there as well.

ygoumaz (approved):
Very nice! This is by far a smoother way to fix the problem.

omichel:
Very good, I would however rename the variable to `HARDWARE_INTERFACE_VERSION` and set it to 3.5 and test if the value is greater or equal to it.

omichel:
More specifically, we could define `HARDWARE_INTERFACE_VERSION_MAJOR` to 3 and `HARDWARE_INTERFACE_VERSION_MINOR` to 5 and test it in a similar way to what is done in https://github.com/cyberbotics/webots/blob/master/src/webots/nodes/WbConnector.cpp#L378

ad-daniel:
as far as I understood we can't do string comparisons in preprocessor statements, that's why I made them on the cmake side and just set the `HARDWARE_INTERFACE_VARIABLE_RENAMING` flag. Rather than checking major+minor I suppose I can check the whole thing at once, but do you suggest I rename the flag to `HARDWARE_INTERFACE_VERSION` anyway to something else?

ad-daniel:
In the first approach I tested I did `set(HARDWARE_INTERFACE_VERSION_MAJOR, hardware_interface_VERSION_MAJOR)` in the CMakeLists.txt and did the comparison on the .cpp side but it didn't work, I think because `set` defines variables as string.

omichel:
Did you try this?
```cmake
add_compile_definitions(HARDWARE_INTERFACE_VERSION_MAJOR =${hardware_interface_VERSION_MAJOR})
add_compile_definitions(HARDWARE_INTERFACE_VERSION_MINOR =${hardware_interface_VERSION_MINOR})
```

ad-daniel:
that does seem to work locally, I'll test

Issue #581, opened 2022-12-29
URDF robot spawning issue on Windows (WSL) and macOS (UTM VM)


ygoumaz:
On windows (WSL) and macOS (UTM VM), spawning URDF robots doesn't always work.

The universal robot multi-robot example works fine, but the following combination of commands fails to spawn the ur5e robot:
```bash
ros2 launch webots_ros2_universal_robot robot_world_launch.py
ros2 launch webots_ros2_universal_robot robot_moveit_nodes_launch.py
```

The path to assets (textures and meshes) seems to be wrong. The path is already modified in the `urdf_spawner.py` script, which allows `ros2 launch webots_ros2_universal_robot multirobot_launch.py` to work fine. A case, where the same method should be applied is probably missing in the workflow.

ygoumaz:
The existing lines for `ros2 launch webots_ros2_universal_robot multirobot_launch.py` to correctly spawn the URDF are here exactly:
https://github.com/cyberbotics/webots_ros2/blob/0d44327ed0b7c06b060fd42b16ec323afc389bc8/webots_ros2_driver/webots_ros2_driver/urdf_spawner.py#L35-L39

PR #610, opened 2023-01-31
Fix URDF spawner on macOS and WSL

### LINKED ISSUE #558: URDFSpawner relative_path_prefix does not work when not using at least the parent directory of the urdf file
**Describe the Bug**
`relative_path_prefix` does not work when not using at least the parent directory of the urdf file. It must at least be the parent directory or one of it's child directory. In a real case scenario it implies that you cannot use `package://my_robot_description_package/urdf/meshes/mymesh.stl` but rather `package://urdf/meshes/mymesh.stl` for your meshes filename, messing with RViz which is not able to find the meshes.

**Steps to Reproduce**
Consider the following architectory :
```
my_robot_description_package/
└── urdf/
    ├── meshes/
    │   └── myrobot.stl
    └── myrobot.urdf
```

and the following urdf : 
```xml
<?xml version="1.0"?>
<robot name="myrobot"
  xmlns:xacro="http://ros.org/wiki/xacro">

<link name="base_footprint">
    <visual>
      <origin xyz="0 0 0" rpy="0 0 0" />
      <geometry>
        <mesh filename="package://my_robot_description_package/urdf/meshes/myrobot.stl" />
      </geometry>
    </visual>

    <collision>
      <origin xyz="0 0 0.0" rpy="0 0 0" />
      <geometry>
        <box size="0.1 0.1 0.1" />
      </geometry>
    </collision>
  </link>

</robot>
```

using the following relative path does not work :
```python
    myrobot_description_package_dir = get_package_share_directory("myrobot_description_package_dir")
    spawm_myrobot = URDFSpawner(
        name="myrobot",
        robot_description=myrobot_description
        relative_path_prefix=myrobot_description_package_dir
    )
```

This line in URDFSpawner throws Errno 2 (No such file or directory) 
```python 
shutil.copytree(relative_path_prefix, os.path.join(container_shared_folder(), os.path.basename(relative_path_prefix)))
```

Here is a workaround I found, using the parent directory of the urdf file as `relative_path_prefix` (it works with child directories too) :
```python
    myrobot_description_package_dir = get_package_share_directory("myrobot_description_package_dir")
    spawm_myrobot = URDFSpawner(
        name="myrobot",
        robot_description=myrobot_description.replace(
            "package://my_robot_description_package/", "package://"
        )
        relative_path_prefix=os.path.join(jrb_description_package_dir, "urdf")
    )
```

**Expected behavior**
Finding a way to make it work without this workaround would be great to ensure having standard urdf filenames and making RViz work.

**Affected Packages**
List of affected packages:
  - webots_ros2_driver
  

**System**
 - Webots Version: R2023a for macOS
 - ROS Version: Humble
 - Operating System: Ubuntu jammy in docker, host macOS Monterey 12.2.1, using local_simulation_server.py

ygoumaz:
Fixes #558 and #581.

## **#581** ##

URDFs can be spawned in 2 different ways:
- By giving the complete robot description and the path to the resources as a prefix to append to assets relative URLs.
- By giving only the path to the URDF file.

The first way was correctly handled on WSL and macOS. The second one was not and is trickier, as the path must be kept to open the file on the Linux side, while the URL should be edited to be compatible with WSL and the shared directory, respectively on Windows and macOS.

To counter this problem, on Windows and macOS, the URDF file is opened by the `ROS2_supervisor` and is passed as the robot description, and the path prefix is set according to the URDF path by searching for the 'share' directory.

The URL edition is now completely done in the supervisor, for clarity.

## **#558** ##

The URLs of xacro files in the universal robot example are reverted to the correct RVIZ format. To fix the issue on macOS, the resource folder is now copied inside another folder that has the same name as the package, to be consistent with the Rviz structure.

### DIFF
--- a/webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
+++ b/webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
@@ -102,8 +102,8 @@ class Ros2Supervisor(Node):
                 # Read the content of the URDF
                 with open(robot.urdf_path, 'r') as file:
                     urdfContent = file.read()
-                if urdfContent is None:
-                    sys.exit('Could not read the URDF file.')
+                    if urdfContent is None:
+                        sys.exit('Could not read the URDF file.')
 
                 # Get the package name and parent resource directory from URDF path
                 split_path = robot.urdf_path.split(os.path.sep)
--- a/webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
+++ b/webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
@@ -19,10 +19,10 @@
 
 
 import os
-import sys
 import re
-import subprocess
 import shutil
+import subprocess
+import sys
 
 import rclpy
 import vehicle
@@ -90,25 +90,30 @@ class Ros2Supervisor(Node):
         box_collision = robot.box_collision if robot.box_collision else False
         init_pos = robot.init_pos if robot.init_pos else None
 
-        # Choose the conversion according to the input + adapt path in function of platform
+        # Choose the conversion according to the input and platform
         if robot.urdf_path:
             if has_shared_folder() or is_wsl():
+                # Check that the file exists and is an URDF
                 if not os.path.isfile(robot.urdf_path):
                     sys.exit('Input file "%s" does not exists.' % robot.urdf_path)
                 if not robot.urdf_path.endswith('.urdf'):
                     sys.exit('"%s" is not a URDF file.' % robot.urdf_path)
 
+                # Read the content of the URDF
                 with open(robot.urdf_path, 'r') as file:
                     urdfContent = file.read()
                 if urdfContent is None:
                     sys.exit('Could not read the URDF file.')
 
+                # Get the package name and parent resource directory from URDF path
                 split_path = robot.urdf_path.split(os.path.sep)
                 for i, folder in (list(enumerate(split_path))):
                     if folder == "share":
                         package_dir = os.path.sep.join(split_path[:i + 2])
                         resource_dir = os.path.sep.join(split_path[:i + 3])
                         break
+                # On macOS, the resources are copied to shared_folder/package_name/resource_folder
+                # The path prefix is updated to the path of the shared folder
                 if has_shared_folder():
                     shared_package_dir = os.path.join(container_shared_folder(), os.path.basename(package_dir))
                     shared_resource_dir = os.path.join(shared_package_dir, os.path.basename(resource_dir))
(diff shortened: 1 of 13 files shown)


webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
```diff
@@ -87,13 +90,77 @@ def __spawn_urdf_robot_callback(self, request, response):
...
-        # Choose the conversion according to the input
+        # Choose the conversion according to the input and platform
         if robot.urdf_path:
-            robot_string = convertUrdfFile(input=robot.urdf_path, robotName=robot_name, normal=normal,
-                                           boxCollision=box_collision, initTranslation=robot_translation,
-                                           initRotation=robot_rotation, initPos=init_pos)
+            if has_shared_folder() or is_wsl():
+                # Check that the file exists and is an URDF
+                if not os.path.isfile(robot.urdf_path):
+                    sys.exit('Input file "%s" does not exists.' % robot.urdf_path)
```
ad-daniel, line 98:
```suggestion
                    sys.exit('Input file "%s" does not exist.' % robot.urdf_path)
```

webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
```diff
@@ -87,13 +90,77 @@ def __spawn_urdf_robot_callback(self, request, response):
...
+                if not robot.urdf_path.endswith('.urdf'):
+                    sys.exit('"%s" is not a URDF file.' % robot.urdf_path)
```
ad-daniel, line 100:
```suggestion
                    sys.exit('"%s" is not an URDF file.' % robot.urdf_path)
```

webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
```diff
@@ -87,13 +90,77 @@ def __spawn_urdf_robot_callback(self, request, response):
...
+
+                # Read the content of the URDF
+                with open(robot.urdf_path, 'r') as file:
+                    urdfContent = file.read()
+                if urdfContent is None:
+                    sys.exit('Could not read the URDF file.')
```
ad-daniel, line 106: this should probably be indented

ad-daniel (changes requested):
Looks good to me, thank you.
@PymZoR may I ask you to check if the proposed patch fixes your issue?

PymZoR:
@ad-daniel I'm going to check it this week-end

ygoumaz:
I will merge this PR, so that we can move on with the release of version 2023.0.2.

@PymZoR feel free to mention if the fix doesn't work for you.

ad-daniel (approved):
Looks good to me

PR #612, opened 2023-02-01
Fix macOS local assets with relative URLs

### LINKED ISSUE #583: Resolving local protos issue in macOS VM
#573 fixed the local protos in WSL by updating the links to the protos in the world file. 

On macOS, the protos folder must first be copied to the shared folder and the links should also be modified.

Currently this is not the case and worlds that use local protos don't work.

ygoumaz:
Fixes #583.

The local assets referenced with relative URL in the world files are copied to the shared folder on macOS. Their URL is also updated to reflect the new relative path to the world.


ad-daniel (approved):
Thank you!

PR #615, opened 2023-02-03
Test CI

ad-daniel:
**Description**

CI jobs are passing but nothing is being tested


ad-daniel:
I've reverted several commits until a time we know for sure things were working and we observe the same behavior, some tests randomly test nothing and the test silently passes. Must be an issue on the ci_industrial side, if it persists we should probably ask in their repo

PR #616, opened 2023-02-03
Optional Ros2 Supervisor

### LINKED ISSUE #509: Ros2Supervisor should be optional
At the moment, a user has to start the `Ros2Supervisor` node even though it doesn't serve any purpose.

E.g.: in the Turtlebot example, the URDF is not used, so the `Ros2Supervisor` node is started just to make the simulation running:
https://github.com/cyberbotics/webots_ros2/blob/3ffa1de47c73180355a88899e81b0d59af37cd19/webots_ros2_turtlebot/launch/robot_launch.py#L111

Furthermore, a robot URL is then necessary:
https://github.com/cyberbotics/webots_ros2/blob/3ffa1de47c73180355a88899e81b0d59af37cd19/webots_ros2_turtlebot/launch/robot_launch.py#L68

For new users, this is not intuitive.

I suggest adding an extra argument in the `WebotsLauncher` (disabled by default):
```py
webots = WebotsLauncher(
  world='world.wbt',
  ros_supervisor=True
)
```
If enabled, ideally, the node should be started implicitly (a user should not worry about starting it).

ygoumaz:
Resolves #509.

The Ros2 supervisor is now optional and is created by the WebotsLauncher Node. The supervisor node can still be used in the launch file if necessary. 

If the Ros2 supervisor is created, it still must figure in the launch description of the launch file, so that it gets started with other nodes.

### DIFF
--- a/webots_ros2/CHANGELOG.rst
+++ b/webots_ros2/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog for package webots_ros2
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+2023.0.3 (2023-XX-XX)
+------------------
+* Ros2Supervisor is now optional.
+
 2023.0.2 (2023-XX-XX)
 ------------------
 * Drop support for Galactic.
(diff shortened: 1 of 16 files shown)


ad-daniel (approved):
Works, thank you

PR #629, opened 2023-02-16
Update sync_controller_lib.sh to copy version.txt

ygoumaz:
With the update of the libController (https://github.com/cyberbotics/webots/pull/5896) to check the version against the one from Webots, the `version.txt` file is needed for compilation.


ad-daniel (approved):
Thank you

ad-daniel:
actually, doesn't master need it too?

ygoumaz:
Currently not, as the change is only on webots/develop. But we could already copy version.txt now in master so that when we release R2023b, we don't need to do anything here. What do you think?

PR #639, opened 2023-02-23
Adding port parameter and usage of webots world project file to launcher.py

sebastianohl:
**Description**
The webots driver launcher do not offer the possibility to set the streaming port. this prevents the use in multi user environments where every user wants to see its own simulation.
Moreover, the launcher only copies the world file and ignores the world project file. Therefore, custom settings (e.g. optional renderings are ignored).

This pull request adds a parameter to webots launcher constructor to set the port. The default parameter setting is set to the default port 1234. Therefore, there should not be any difference for existing setups.
Moreover, the copy of the world project file is added. The execute()-Function searches if a .wbproj file with the same name as the world file exists and then copies it to the location of the temporary world file. When webots starts, it uses the world project file.

**Related Issues**

**Affected Packages**
  - webots_ros2_driver

**Tasks**

### DIFF
--- a/webots_ros2/CHANGELOG.rst
+++ b/webots_ros2/CHANGELOG.rst
@@ -8,6 +8,8 @@ Changelog for package webots_ros2
 * Fixed the spawn of URDF robots in WSL and macOS when using full path.
 * Fixed relative assets in macOS.
 * Ros2Supervisor is now optional.
+* Adding port, stream type parameters to webots_laucher
+* Copying .wbproj when launching a webots world via webots_launcher
 
 2023.0.1 (2023-01-05)
 ------------------
--- a/webots_ros2/CHANGELOG.rst
+++ b/webots_ros2/CHANGELOG.rst
@@ -2,14 +2,17 @@
 Changelog for package webots_ros2
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+2023.0.3 (2023-XX-XX)
+------------------
+* Adding port, stream type parameters to webots_laucher
+* Copying .wbproj when launching a Webots world via webots_launcher
+
 2023.0.2 (2023-02-07)
 ------------------
 * Drop support for Galactic.
 * Fixed the spawn of URDF robots in WSL and macOS when using full path.
 * Fixed relative assets in macOS.
 * Ros2Supervisor is now optional.
-* Adding port, stream type parameters to webots_laucher
-* Copying .wbproj when launching a webots world via webots_launcher
 
 2023.0.1 (2023-01-05)
 ------------------
--- a/webots_ros2_driver/CHANGELOG.rst
+++ b/webots_ros2_driver/CHANGELOG.rst
@@ -1,6 +1,10 @@
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Changelog for package webots_ros2_driver
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+2023.0.3
+------------------
+* Adding port, stream type parameters to webots_laucher
+* Copying .wbproj when launching a webots world via webots_launcher
 
 2023.0.2 (2023-02-07)
 ------------------
--- a/webots_ros2_driver/CHANGELOG.rst
+++ b/webots_ros2_driver/CHANGELOG.rst
@@ -1,7 +1,8 @@
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Changelog for package webots_ros2_driver
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-2023.0.3
+2023.0.3 (2023-XX-XX)
 ------------------
 * Adding port, stream type parameters to webots_laucher
 * Copying .wbproj when launching a webots world via webots_launcher
--- a/webots_ros2_driver/webots_ros2_driver/utils.py
+++ b/webots_ros2_driver/webots_ros2_driver/utils.py
@@ -135,9 +135,9 @@ def get_host_ip():
         sys.exit('Unable to get host IP address. \'ip route\' could not be executed.')
 
 
-def controller_url_prefix():
+def controller_url_prefix(port="1234"):
     if has_shared_folder() or is_wsl():
-        return 'tcp://' + (get_host_ip() if has_shared_folder() else get_wsl_ip_address()) + ':1234/'
+        return 'tcp://' + (get_host_ip() if has_shared_folder() else get_wsl_ip_address()) + ':' + port + '/'
     else:
         return ''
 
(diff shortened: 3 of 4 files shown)


webots_ros2_driver/webots_ros2_driver/webots_launcher.py
```diff
@@ -155,6 +160,17 @@ def execute(self, context: LaunchContext):
 
         shutil.copy2(world_path, self.__world_copy.name)
 
+        # look for a wbproj file and copy if available
+        wbproj_path = Path(world_path)
+        wbproj_path = wbproj_path.with_suffix('.wbproj')
+        wbproj_path = wbproj_path.with_name("."+wbproj_path.name)
```
ygoumaz, line 166:
```suggestion
        wbproj_path = Path(world_path).with_name('.' + Path(world_path).stem + '.wbproj')
```

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
```diff
@@ -155,6 +160,17 @@ def execute(self, context: LaunchContext):
...
+
+        if wbproj_path.exists():
+            wbproj_copy_path = Path(self.__world_copy.name)
+            wbproj_copy_path = wbproj_copy_path.with_suffix('.wbproj')
+            wbproj_copy_path = wbproj_copy_path.with_name("."+wbproj_copy_path.name)
```
ygoumaz, line 171:
```suggestion
            wbproj_copy_path = Path(self.__world_copy.name).with_name('.' + Path(self.__world_copy.name).stem + '.wbproj')
```

ygoumaz (changes requested):
Very nice contribution, thanks a lot! I have some minor suggestions on the code. Also, two additional steps should be added:

- On macOS, resources should be copied to the shared folder. The wbproj should also be copied so that macOS users also benefit from the feature. This can be added here along with the world file: https://github.com/cyberbotics/webots_ros2/blob/801c73c7c4e24361f73f2c532542f75aadcc5340/webots_ros2_driver/webots_ros2_driver/webots_launcher.py#L213
- We now are able to choose the port of Webots but we should also be able to easier choose the target port of started controllers/nodes. Based on how we setup the target in the examples of the packages, the `controller_url_prefix()` function should be enhanced with an additional `port` parameter. The default value should be 1234 as well if no port is given. https://github.com/cyberbotics/webots_ros2/blob/801c73c7c4e24361f73f2c532542f75aadcc5340/webots_ros2_driver/webots_ros2_driver/utils.py#L138

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
sebastianohl: I added the changed in the new commit

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
sebastianohl: I added the changed in the new commit

ygoumaz (changes requested):
Nice catch for the supervisor! I would just add the `port-argument` as argument when creating the `Ros2SupervisorLauncher` object, so that it is effectively connecting to the correct instance of Webots: https://github.com/cyberbotics/webots_ros2/blob/d5a987829fdcfde7456a41cdfdbb61fffcaac112/webots_ros2_driver/webots_ros2_driver/webots_launcher.py#L67

Finally, you should add the changes to the [changelog](https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_driver/CHANGELOG.rst) of `webots_ros2_driver` and in the [main changelog](https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2/CHANGELOG.rst) (you can add them under version `2023.0.3`).

webots_ros2_driver/webots_ros2_driver/utils.py
```diff
@@ -135,9 +135,9 @@ def get_host_ip():
         sys.exit('Unable to get host IP address. \'ip route\' could not be executed.')
 
 
-def controller_url_prefix():
+def controller_url_prefix(port="1234"):
```
omichel, line 138:
```suggestion
def controller_url_prefix(port='1234'):
```

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
```diff
@@ -229,15 +245,15 @@ def _shutdown_process(self, context, *, send_sigint):
 
 
 class Ros2SupervisorLauncher(Node):
-    def __init__(self, output='screen', respawn=True, **kwargs):
+    def __init__(self, output='screen', respawn=True, port="1234", **kwargs):
```
omichel, line 248:
```suggestion
    def __init__(self, output='screen', respawn=True, port='1234', **kwargs):
```

omichel (changes requested):
We try to prefer single quotes to double quotes.

webots_ros2_driver/CHANGELOG.rst
```diff
@@ -1,6 +1,10 @@
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Changelog for package webots_ros2_driver
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+2023.0.3
+------------------
+* Adding port, stream type parameters to webots_laucher
+* Copying .wbproj when launching a webots world via webots_launcher
```
ygoumaz, line 7:
```suggestion
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

2023.0.3 (2023-XX-XX)
------------------
* Adding port, stream type parameters to webots_laucher
* Copying .wbproj when launching a webots world via webots_launcher
```

webots_ros2/CHANGELOG.rst
```diff
@@ -8,6 +8,8 @@ Changelog for package webots_ros2
 * Fixed the spawn of URDF robots in WSL and macOS when using full path.
 * Fixed relative assets in macOS.
 * Ros2Supervisor is now optional.
+* Adding port, stream type parameters to webots_laucher
+* Copying .wbproj when launching a webots world via webots_launcher
```
ygoumaz, line 12: This should be moved to a new 2023.0.3 (2023-XX-XX) section.

ygoumaz (changes requested):
Thank you. I just have a few minor comments on the changelogs.

webots_ros2_driver/webots_ros2_driver/webots_launcher.py
```diff
@@ -52,7 +53,8 @@ def perform(self, context):
 
 
 class WebotsLauncher(ExecuteProcess):
-    def __init__(self, output='screen', world=None, gui=True, mode='realtime', stream=False, ros2_supervisor=False, **kwargs):
+    def __init__(self, output='screen', world=None, gui=True, mode='realtime', stream=False, ros2_supervisor=False,
+                 port="1234", **kwargs):
```
ygoumaz, line 57:
```suggestion
                 port='1234', **kwargs):
```

ygoumaz (changes requested):
I spotted an additional double quote.

ygoumaz (approved):
All good to me now! Thank you!

PR #640, opened 2023-02-23
Temporarily fix `diff_drive controller` bad variable name

### LINKED ISSUE #593: MacOS Compatibility Issues with RViz2 and the '/tf' Topic
On macOS, RViz2 can be launched in the VM, but it encounters an error with the `/tf` topic.

`robot_state_publisher` publishes data on this topic. The data type contains a field `frame_id` and a field `child_frame_id`, which should obviously be different. On Linux and WSL, this is correct and RViz is able to use the data of the topic. On macOS, `frame_id` and `child_fame_id` are very often the same.

The only difference is the fact that Webots is started remotely, which may introduce some latency/synchronization issues.

ygoumaz:
Fix #593 caused by an update of `diff_drive_controller`.

This can be reverted once version `2.16.0` of the package is available.

### DIFF
--- a/webots_ros2_epuck/resource/ros2_control.yml
+++ b/webots_ros2_epuck/resource/ros2_control.yml
@@ -17,3 +17,4 @@ diffdrive_controller:
     wheel_radius: 0.02
 
     use_stamped_vel: false
+    base_frame_id: "base_link"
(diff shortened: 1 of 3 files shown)


omichel (approved):
Thank you.

PR #642, opened 2023-02-24
Improve Tiago navigation

ygoumaz:
RViz and the navigation must be started once the Webots simulation is ready to avoid a lot of warnings in the terminal.

Also the test should publish the initial position only when a subscriber is ready. In addition, the current succeeding condition of the test is wrong and should be `< -2.0` for the robot to move a few seconds towards the given goal position.

The TIAGo test is skipped, because the Navigation 2 package is outdated and is not working with the given configuration.

### DIFF
--- a/scripts/ci_after_init.bash
+++ b/scripts/ci_after_init.bash
@@ -26,7 +26,6 @@ if [[ $(lsb_release -rs) == "22.04" && ${WEBOTS_RELEASE_VERSION} == "2022a" ]]; 
   mv /tmp/openssl-1.1/* /usr/local/webots/lib/webots/
 fi
 
-apt install -y ros-${ROS_DISTRO}-ros2topic
 # The following packages are not available in the ROS 2 Rolling distribution. Therefore, we cannot include them in the package.xml, but we have to install them manually here.
 if [[ "${ROS_DISTRO}" != "rolling" ]]; then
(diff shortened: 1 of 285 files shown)


ygoumaz:
The PR is (finally) ready for review!

ygoumaz:
Back to draft...

ad-daniel (approved):
Thank you. By the way I still have a doubt concerning whether the CI is actually working, the timings seem correct but the summary line at the end now shows 
`Summary: 436 tests, 0 errors, 0 failures, 127 skipped`
If you go back like 2 months in the test suite, there weren't as many being skipped. Do you have any idea why?

ad-daniel (approved):

ygoumaz:
I think this is normal because in addition to our tests, there is also other tests performed from other ament packages. I believe they added more tests in these packages in the last weeks. These updates may even be a reason why CI industrial was not working properly a few days ago. 

If we check the logs in detail, we see that our tests succeed and are skipped when needed (turtlebot for example).

ad-daniel:
Those I saw had significantly less skipped, for instance [this](https://github.com/cyberbotics/webots_ros2/actions/runs/3898617271/jobs/6657517130) from January 12th 
```
Summary: 457 tests, 0 errors, 0 failures, 3 skipped
```

ygoumaz:
In fact, the high number of tests skipped is only on Humble. On Foxy, less tests are skipped. But this seems consistent between the test you mention and two ones from today.

PR #658, opened 2023-03-04
Fix ros2_control wheel_separation param for turtlebot3

ftyghome:
**Description**
The wheel_separation param in ros2control.yml is incorrect, which would make the odometry not accurate.
While the wheel_separation is set to 1.60, when making the robot to turn and the odometry records 360 degrees, the orientation of the robot in webots is not matched with the odometry.

The initial orientation is shown in the picture below:
image

After making a 360 degree turn in odometry, the orientation is shown below:
Screenshot from 2023-03-04 19-05-55

I found that the cause of the problem is related to the wheel_separation parameter. In the "TurtleBot3Burger.proto" file, the position of the left joint is "0 0.08 0.033", and the right joint is "0 -0.08 0.033". So it seems that the separation is 0.16. But that does not take the wheel cylinder into account. The height of the cylinder is 0.018, so the actual wheel separation should be 0.16+2*(0.018/2) = 0.178.

Changing the param would make the odometry works accurately.

### DIFF
--- a/webots_ros2_turtlebot/resource/ros2control.yml
+++ b/webots_ros2_turtlebot/resource/ros2control.yml
@@ -13,7 +13,7 @@ diffdrive_controller:
     left_wheel_names: ["left wheel motor"]
     right_wheel_names: ["right wheel motor"]
 
-    wheel_separation: 0.160
+    wheel_separation: 0.178
     wheel_radius: 0.033
 
     use_stamped_vel: false


omichel (approved):
Thank you for this nicely documented contribution.

ftyghome:
> Thank you for this nicely documented contribution. Please note however, that the position of the wheels as indicated in the [endPoint translation](https://github.com/cyberbotics/webots/blob/master/projects/robots/robotis/turtlebot/protos/TurtleBot3Burger.proto#L57) defines the center of the wheel [cylinder defined in the bounding object](https://github.com/cyberbotics/webots/blob/master/projects/robots/robotis/turtlebot/protos/TurtleBot3Burger.proto#L113-L116). Therefore, I believe the value of 0.16 is correct to define the distance between the centers of the two wheels. The value you computed is actually the distance between the outer parts of the wheels. Is it what you need to input to the `diffdrive` controller?

May be it is. I don't know how webots's physics system works, but the experiment result shows that I need to input the distance between the outer parts of the wheels, then the odometry would be accurate.

omichel:
> May be it is. I don't know how webots's physics system works, but the experiment result shows that I need to input the distance between the outer parts of the wheels, then the odometry would be accurate.

OK, it sounds a bit strange to me, but it may me fine if that gives better results for the odometry.

PR #664, opened 2023-03-08
adding functionally to copy a sumo network to the tmp folder with the right name

sebastianohl:
**Description**
If using the webots launcher for a world file containing a SUMO traffic simulation, the sumo network folder is not copied to the temp folder. Therefore, the traffic simulation does not start. It is possible to give absolute path in the SumoInterface node. However, this does not work if the folders are located in home-directories and used by different users.

The proposed PR adds a parameter to the WebotsLauncher constructor, requesting to also copy the sumo network folder to the new world file location. It is set to False by default, so no existing code needs to be changed.

**Affected Packages**
List of affected packages:
  - webots_ros2_driver


omichel:
Thank you for this contribution. However, I believe we could make it simpler: copy the sumo network file only if it exists and remove the `sumo` option from the constructor of the launcher, so that users don't have to worry about it.

omichel (approved):
Great, thank you.

PR #672, opened 2023-03-13
Fix random Nav2 failure with Fast DDS

ygoumaz:
Nav2 random failures in the TIAGo tests are due to issues with the default Fast DDS on Humble. See https://github.com/ros-planning/navigation2/issues/3033#issuecomment-1457854463 comment and issue.

We switch the DDS to Cyclone to fix the problem.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       AFTER_INIT: ./scripts/ci_after_init.bash ${ROS_DISTRO} ${ROS_REPO}
-      BEFORE_INIT_EMBED: source scripts/ci_before_init_embed.bash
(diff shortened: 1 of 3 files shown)


ad-daniel (changes requested):
Shouldn't rolling require the same patch? It doesn't appear to be a solved yet based on the discussion you linked, so that distro should have the same issue no?

Also might be a good idea to leave a comment that references this PR or the linked one for future reference, as this fix only bypasses the issue, and therefore I assume that if somebody installs webots_ros2 locally he might face the same problems since these bash scripts aren't run locally

ad-daniel (approved):
Thank you

ygoumaz:
If the error is completely fixed in the future, we may remove this fix. Also I have added [a section](https://github.com/cyberbotics/webots_ros2/wiki/Example-TIAGo#known-issue-nav2-and-fast-dds) in the TIAGo instuctions so that users that encounter the issue locally can use the temporary solution.

PR #676, opened 2023-03-15
Calibrate Tiago odometry and restore SLAM-Toolbox

### LINKED ISSUE #635: Drop slam-toolbox for other more robust packages
The slam-toolbox package is used in the [Tiago tutorial](https://github.com/cyberbotics/webots_ros2/wiki/Example-TIAGo) for SLAM. There have been reports that the package can be inaccurate when used with the diff-drive controller (see https://github.com/cyberbotics/webots/issues/5540#issuecomment-1371748895). Here is the kind of map generated with the Tiago:

Screenshot from 2023-02-20 17-23-04

As such, I suggest to replace the package with a more accurate alternative like AMCL or Cartographer in the Tiago tutorial. Cartographer is already used in the [Turtlebot tutorial](https://github.com/cyberbotics/webots_ros2/wiki/SLAM-with-TurtleBot3). @delihus has also demonstrated that AMCL gives better results, in the scope of a project using the ROSbot.

The plan is to transfer tutorials to the official ROS 2 documentation. If there is no opposition, slam-toolbox will be replaced by another package in the new Tiago tutorial.

ygoumaz:
Resolve #635.

The TIAGo odometry is calibrated to allow a better result with SLAM toolbox. 

However, the SLAM is still not 100% accurate, as the `scan_matching` option is not working correctly and is disabled in the calibration file.

The user can now choose between SLAM-toolbox and cartographer when running the TIAGo example.


ad-daniel (changes requested):
I'm unable to launch it on foxy with `ros2 launch webots_ros2_tiago robot_launch.py rviz:=true slam:=true`, I get a ton of:
```
[rviz2-8] Parsing robot urdf xml string.
[rviz2-8] Warning: Invalid frame ID "map" passed to canTransform argument target_frame - frame does not exist
[rviz2-8]          at line 133 in /tmp/binarydeb/ros-foxy-tf2-0.13.13/src/buffer_core.cpp
[rviz2-8] Warning: Invalid frame ID "map" passed to canTransform argument target_frame - frame does not exist
[rviz2-8]          at line 133 in /tmp/binarydeb/ros-foxy-tf2-0.13.13/src/buffer_core.cpp
[rviz2-8] Warning: Invalid frame ID "map" passed to canTransform argument target_frame - frame does not exist
[rviz2-8]          at line 133 in /tmp/binarydeb/ros-foxy-tf2-0.13.13/src/buffer_core.cpp

```

ygoumaz:
slam option doesn't exist anymore. It is slam_toolbox or slam_cartographer now. This allows to choose between both packages.

ad-daniel (dismissed):
> slam option doesn't exist anymore. It is slam_toolbox or slam_cartographer now.

I see, works now. We should update the wiki (or tutorials if they've been ported) when this is merged then

ygoumaz:
I just pushed a modification to keep track of the weird dimensions in the parameters

ad-daniel (approved):

webots_ros2_tiago/resource/ros2_control.yml
```diff
@@ -13,8 +13,9 @@ diffdrive_controller:
     left_wheel_names: ["wheel_left_joint"]
     right_wheel_names: ["wheel_right_joint"]
 
-    wheel_separation: 0.5
-    wheel_radius: 0.12
+    # Values don't correspond exactly to the ones in the robot URDF but result in more accurate odometry
```
delihus, line 16: For this change I think you should use these parameters:
```yaml
wheel_separation_multiplier: 0.88
left_wheel_radius_multiplier: 0.8167
right_wheel_radius_multiplier: 0.8167
```
The wheel radius and the  wheel separation are physical parameters which should be the same as in the URDF.

omichel: Thank you. @ygoumaz is going to investigate this soon.

PR #656, opened 2023-03-03
Adding receiver emitter nodes

angel-ayala:
Hi,

I'm currently following the [Tutorial Creating a Custom Cpp Plugin](https://github.com/cyberbotics/webots_ros2/wiki/Tutorial-Creating-a-Custom-Cpp-Plugin) to include the [Emitter](https://www.cyberbotics.com/doc/reference/emitter?tab-language=c) and [Receiver](https://www.cyberbotics.com/doc/reference/receiver?tab-language=c). nodes.

This added:
- Two plugins to `webots_ros2_driver` module: `Ros2Emitter` and `Ros2Receiver`.
- One `StringStamped` message and `SetString` service to `webots_ros2_msgs`.

Current dev-stage allows:
-  the `Receiver` node publish in a `StringStamped` topic  `/data` if `wb_receiver_get_queue_length(mReceiver) > 0` the results of `wb_receiver_get_data(mReceiver)` .
- the `Emitter` node create a `SetString` service `/send` to perform the `wb_emitter_send()` function.
Stills require to implement all the remaining functions.

**TODO**
  - [ ] `Emitter` node's missing functions.
  - [ ] `Receiver` node's missing functions.
  - [ ] perform tests.

**Additional context**
I'm currently developing another package in order to use these plugins simulating a communication between two radio frequency controllers.
I have a `RCController` which controls a `RadioController` node. 
There are two `RadioController` nodes in the scene: `RCControlMain` and `RCControlSlave`

When I launch the `robot_launch.py` file, seems that the  `RCControlMain` keeps waiting for the `/send` service to start.
The expected behavior is a continuous string sharing between both `RadioController`, but seems that one is being replaced by the other.

Any thoughts?

Here is the console's output:
```
$ ros2 launch rc_comm robot_launch.py 
[INFO] [launch]: All log files can be found below /home/kelo/.ros/log/2023-03-03-15-25-53-387733-mjro-linux-hp-66077
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [webots-1]: process started with pid [66078]
[INFO] [ros2_supervisor.py-2]: process started with pid [66080]
[INFO] [driver-3]: process started with pid [66082]
[INFO] [driver-4]: process started with pid [66085]
[driver-3] Cannot connect to Webots instance, retrying for another 50 seconds...
[driver-4] Cannot connect to Webots instance, retrying for another 50 seconds...
[...]

PR #677, opened 2023-03-17
Adding Compass node

angel-ayala:
**Description**
Adding the [Compass node](https://www.cyberbotics.com/doc/reference/compass) to be used as a device

**Affected Packages**
List of affected packages:
  - webots_ros2_driver
  - webots_ros2_msgs


ygoumaz (changes requested):
Thanks for this contribution! I have some comments on the code and implementation. In addition, an entry for the enhancement should be added to the [main webots_ros2 changelog](https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2/CHANGELOG.rst) and [webots_ros2_driver changelog](https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_driver/CHANGELOG.rst).

webots_ros2_driver/src/plugins/static/Ros2Compass.cpp
```diff
@@ -0,0 +1,99 @@
...
+  }
+  void Ros2Compass::publishData() {
+    const double *compass_values = wb_compass_get_values(mCompass);
+    mDataMessage.header.stamp = mNode->get_clock()->now();
+    mDataMessage.magnetic_field.x = compass_values[0];
+    mDataMessage.magnetic_field.y = compass_values[1];
+    mDataMessage.magnetic_field.z = compass_values[2];
+    mDataPublisher->publish(mDataMessage);
+  }
+  void Ros2Compass::enable_callback(const std::shared_ptr<webots_ros2_msgs::srv::SetInt::Request> request,
```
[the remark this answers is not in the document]
angel-ayala: I followed the definition of the ROS language for the node here https://www.cyberbotics.com/doc/reference/compass?tab-language=ros#wb_compass_enable
In regards to the timestep value, this is readed from the URDF file definition by the [Ros2SensorPlugin.cpp file](https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_driver/src/plugins/Ros2SensorPlugin.cpp) instantiating the `mPublishTimestepSyncedMs` var.

webots_ros2_driver/src/plugins/static/Ros2Compass.cpp
```diff
@@ -0,0 +1,99 @@
...
+      mAlwaysOn = false;
+      mIsEnabled = false;
+    }
+    response->success = true;
+  }
+  void Ros2Compass::get_sampling_period_callback(const std::shared_ptr<webots_ros2_msgs::srv::GetInt::Request> request,
+                                                 std::shared_ptr<webots_ros2_msgs::srv::GetInt::Response> response) {
+    response->value = request->ask ? wb_compass_get_sampling_period(mCompass) : 0;
+  }
+  void Ros2Compass::get_lookup_table_callback(const std::shared_ptr<webots_ros2_msgs::srv::GetFloatArray::Request> request,
```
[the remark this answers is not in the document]
angel-ayala: Yes, you are right, I will update those lines including the full size of the vector using `wb_compass_get_lookup_table_size()`

webots_ros2_driver/src/plugins/static/Ros2Compass.cpp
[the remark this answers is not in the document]
ygoumaz: The ROS definition in the Webots documentation is for ROS 1 and not ROS 2. However, I think you are doing well by adding these services to your plugin. I would add a new service called `mDisableService` that allows to disable separately the device, so that it is consistent with the rest of the API.

In the future, we should also add equivalent services to other existing plugin devices of webots_ros2_driver (gps, camera, etc.)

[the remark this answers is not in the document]
angel-ayala: Sure, seems ok for me. I will add this discussed changes in the following commit.

webots_ros2_driver/src/plugins/static/Ros2Compass.cpp
```diff
@@ -0,0 +1,110 @@
...
+    if (request->value) {
+      wb_compass_disable(mCompass);
+      mAlwaysOn = false;
+      mIsEnabled = false;
+    }
+    response->success = true;
+  }
+  void Ros2Compass::get_sampling_period_callback(const std::shared_ptr<webots_ros2_msgs::srv::GetInt::Request> request,
+                                                 std::shared_ptr<webots_ros2_msgs::srv::GetInt::Response> response) {
+    response->value = request->ask ? wb_compass_get_sampling_period(mCompass) : 0;
```
ygoumaz, line 93: No need to use `request->ask`. `GetInt` message doesn't need to have an 'ask' input argument. Moreover, `wb_compass_get_sampling_period` returns 0 if the compass is not enabled.

webots_ros2_driver/src/plugins/static/Ros2Compass.cpp
```diff
@@ -0,0 +1,110 @@
...
+    if (request->ask) {
+      const double *lookup_table = wb_compass_get_lookup_table(mCompass);
+      for (int i = 0; i < table_size; i++) {
+        int offset = i * 3;
+        values.at(offset) = lookup_table[offset];
+        values.at(offset + 1) = lookup_table[offset + 1];
+        values.at(offset + 2) = lookup_table[offset + 2];
+      }
+    }
+    response->values = values;
```
ygoumaz, line 108:
```suggestion
    const double *lookup_table = wb_compass_get_lookup_table(mCompass);
    for (int i = 0; i < table_size; i++) {
      int offset = i * 3;
      values.at(offset) = lookup_table[offset];
      values.at(offset + 1) = lookup_table[offset + 1];
      values.at(offset + 2) = lookup_table[offset + 2];
    }
    response->value = values;
```

We can get rid of the ask input. This is not mandatory.

webots_ros2_driver/src/plugins/static/Ros2Compass.cpp
```diff
@@ -0,0 +1,110 @@
...
+  void Ros2Compass::enable_callback(const std::shared_ptr<webots_ros2_msgs::srv::SetInt::Request> request,
+                                    std::shared_ptr<webots_ros2_msgs::srv::SetInt::Response> response) {
+    if (request->value) {
+      wb_compass_enable(mCompass, mPublishTimestepSyncedMs);
+      mAlwaysOn = true;
+      mIsEnabled = true;
+    }
+    response->success = true;
+  }
+  void Ros2Compass::disable_callback(const std::shared_ptr<webots_ros2_msgs::srv::SetInt::Request> request,
```
ygoumaz, line 82: This service could just be of type `GetBool`. No input value is required and it should return true when done.

ygoumaz (changes requested):
I have some additional requests about services. The changes introduced for the `Emitter` and `Receiver` in #656 were not correct. Services messages that are used to get values (like `GetInt` , `GetFloat` and `GetBool`) should not contain any input (e.g. `bool ask`). Services with no input arguments are totally fine.

Additionally, we should add a service to get the lookup table size to be consistent with other languages of the API.

Emitter and Receiver files should be fixed in another PR, but `.srv` files can be fixed in this PR.

Changelog files should also be filled.

webots_ros2_driver/src/plugins/static/Ros2Compass.cpp
```diff
@@ -0,0 +1,110 @@
...
+    mDataPublisher->publish(mDataMessage);
+  }
+  void Ros2Compass::enable_callback(const std::shared_ptr<webots_ros2_msgs::srv::SetInt::Request> request,
+                                    std::shared_ptr<webots_ros2_msgs::srv::SetInt::Response> response) {
+    if (request->value) {
+      wb_compass_enable(mCompass, mPublishTimestepSyncedMs);
+      mAlwaysOn = true;
+      mIsEnabled = true;
+    }
+    response->success = true;
```
angel-ayala, line 80: After the last comments I should also change this lines right? using `request->value` as the time step argument, right?
```suggestion
   wb_compass_enable(mCompass, request->value);
   mAlwaysOn = true;
   mIsEnabled = true;
   response->success = true;
```

webots_ros2_driver/src/plugins/static/Ros2Compass.cpp
```diff
@@ -0,0 +1,104 @@
...
+    mAlwaysOn = true;
+    mIsEnabled = true;
+    response->success = true;
+  }
+  void Ros2Compass::disable_callback(const std::shared_ptr<webots_ros2_msgs::srv::GetBool::Request> request,
+                                     std::shared_ptr<webots_ros2_msgs::srv::GetBool::Response> response) {
+    wb_compass_disable(mCompass);
+    mAlwaysOn = false;
+    mIsEnabled = false;
+    response->success = true;
```
angel-ayala, line 85:
```suggestion
    response->value = true;
```

webots_ros2_driver/src/plugins/static/Ros2Compass.cpp
ygoumaz: Yes, you are right. I missed this one. The requested value should be used and not the global timestep.

webots_ros2_msgs/srv/GetBool.srv
```diff
@@ -0,0 +1,2 @@
+---
+bool value
```
ygoumaz, line 2:
```suggestion
bool value

```

webots_ros2_msgs/srv/GetFloatArray.srv
```diff
@@ -0,0 +1,2 @@
+---
+float64[] values
```
ygoumaz, line 2:
```suggestion
float64[] value
```

As a convention, we keep it singular.

webots_ros2_driver/src/plugins/static/Ros2Compass.cpp
```diff
@@ -0,0 +1,104 @@
...
+    int table_size = wb_compass_get_lookup_table_size(mCompass);
+    std::vector<double> values(table_size * 3);
+    const double *lookup_table = wb_compass_get_lookup_table(mCompass);
+    for (int i = 0; i < table_size; i++) {
+      int offset = i * 3;
+      values.at(offset) = lookup_table[offset];
+      values.at(offset + 1) = lookup_table[offset + 1];
+      values.at(offset + 2) = lookup_table[offset + 2];
+    }
+    response->values = values;
```
ygoumaz, line 102:
```suggestion
    response->value = values;
```

ygoumaz (changes requested):
A few minor suggestions left. Also as mentioned in my last review, an additional service for the lookup table size should be added.

webots_ros2_driver/src/plugins/static/Ros2Compass.cpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+    return mIsEnabled;
+  }
+  void Ros2Compass::publishData() {
+    const double *compass_values = wb_compass_get_values(mCompass);
+    mDataMessage.header.stamp = mNode->get_clock()->now();
+    mDataMessage.magnetic_field.x = compass_values[0];
+    mDataMessage.magnetic_field.y = compass_values[1];
+    mDataMessage.magnetic_field.z = compass_values[2];
+    mDataPublisher->publish(mDataMessage);
```
lukicdarkoo, line 71: Sorry for randomly jumping into the PR, but a few notes here.

According to the documentation, these values should be in Tesla units:

> x, y, and z components of the field vector in Tesla
- http://docs.ros.org/en/noetic/api/sensor_msgs/html/msg/MagneticField.html

Therefore, we should take the raw value from Webots, calculate the actual value using the lookup table and convert those values to Teslas. As the result, the `get_lookup_table` service is irrelevant.

Other services `enable`, `disable`, and `get_sampling_period` are almost never used in real hardware. Therefore, exposing them could potentially make the sim2real transfer harder (we want to minimize the reality gap).

It would be probably useful to publish other types of messages, e.g.:
https://github.com/clearpathrobotics/imu_compass/blob/master/src/imu_compass.cpp
Maybe `sensor_msgs/Imu` with only the `orientation` field populated. Ideally, we should take a look into a few existing compass interfaces and expose only the most generic part of their ROS API.

ygoumaz: Thanks for your comments! They are really useful. I now understand why all these services are not included in other plugins. I agree that these services should therefore not be included here either. This also applies to the new Emitter and Receiver nodes. I also agree that we should be consistent with other compass interfaces, like the one you have linked.

ygoumaz: Actually, the Webots Compass node (and magnetometers in general) return a 3D vector of the North direction. In the interface linked by @lukicdarkoo, the code gets the compass measuremement in a message of type [Vector3Stamped](http://docs.ros.org/en/noetic/api/geometry_msgs/html/msg/Vector3Stamped.html). We should use this instead of the MagneticField type.

lukicdarkoo: @ygoumaz I would probably find Vector3Stamped more useful than MagneticField. Also, a heading (as Float32) would be quite useful for most use cases.

ygoumaz:
Yes, sure! Here are the main points to modify in the new Compass node, and in the existing Emitter and Receiver ones:
- All services should be removed. As mentioned by @lukicdarkoo, they don't reflect the reality, so we don't want to add them. Also, this is why they are not implemented in other sensors plugins (GPS, etc.).
- Two separate Compass data should be published to topics:
  - The first one is the vector returned by `wb_compass_get_values() `and should be published to the topic you have already created. But instead of `MagneticField` messages type, it should be `Vector3Stamped` type.
  - The second one is not yet implemented but should by a Floating point value in degrees (or radians) representing the heading of the robot. The computation can be achieved using a simple algorithm (you should implement it in Python):
    ```c
    double get_bearing_in_degrees() {
      const double *north = wb_compass_get_values(tag);
      double rad = atan2(north[1], north[0]);
      double bearing = (rad - 1.5708) / M_PI * 180.0;
      if (bearing < 0.0)
        bearing = bearing + 360.0;
      return bearing;
    }
    ```
- All unnecessary services messages (.srv) should be removed.

The respective changes about the Compass should be performed in this PR, when changes on the Emitter and Receiver should be performed in #680.

ygoumaz:
>   * The second one is not yet implemented but should be a Floating point value in degrees (or radians) representing the heading of the robot. The computation can be achieved using a simple algorithm (you should implement it in Python)

I am sorry, there is obviously an error in my last comment. The algorithm should not be implemented in Python. It can simply be added to the cpp plugin file of the Compass.

Once the PR is ready for a new review, you can re-request it by using the double arrow symbol on the top right of the page.

webots_ros2_driver/src/plugins/static/Ros2Compass.cpp
```diff
@@ -0,0 +1,77 @@
...
+  void Ros2Compass::publishData() {
+    const double *north_vector = wb_compass_get_values(mCompass);
+    mVectorMessage.header.stamp = mNode->get_clock()->now();
+    mVectorMessage.vector.x = north_vector[0];
+    mVectorMessage.vector.y = north_vector[1];
+    mVectorMessage.vector.z = north_vector[2];
+    mVectorPublisher->publish(mVectorMessage);
+
+    // north degree
+    double rad = atan2(north_vector[1], north_vector[0]);
```
ygoumaz, line 68:
```suggestion
    const double rad = atan2(north_vector[1], north_vector[0]);
```

webots_ros2_msgs/srv/GetFloatArray.srv
```diff
@@ -0,0 +1,3 @@
+bool ask
+---
+float64[] values
```
ygoumaz, line 3:
```suggestion
```

webots_ros2_msgs/CMakeLists.txt
```diff
@@ -23,13 +23,15 @@ find_package(std_msgs REQUIRED)
...
 set(msg_files
+  "msg/FloatStamped.msg"
   "msg/StringStamped.msg"
   "msg/CameraRecognitionObject.msg"
   "msg/CameraRecognitionObjects.msg"
   "msg/UrdfRobot.msg"
 )
 set(srv_files
   "srv/GetFloat.srv"
+  "srv/GetFloatArray.srv"
```
ygoumaz, line 34:
```suggestion
```

ygoumaz (changes requested):
Some minor remaining comments.

ygoumaz (changes requested):
[GetFloatArray.srv](https://github.com/cyberbotics/webots_ros2/blob/446b7c15f1e3c8e0b9768f52bba3653a98b72479/webots_ros2_msgs/srv/GetFloatArray.srv) should be completely deleted.

ygoumaz:
I have tested the compass on the mavic example and it works nicely. This should be the last change to do before merging this PR!

ygoumaz (approved):
Thank you. The PR is ready to be merged. I just posted an additional minor suggestion. Feel free to apply it before merging. I believe using `north_vector` as topic name is more explicit than `values`.

ygoumaz (approved):

ygoumaz:
I agree for some last changes. But a lot of changes appeared after other previous changes. Also I usually make a first code review and then test the compilation and behavior. Certain changes may get my attention only after these steps.

ygoumaz (approved):

PR #683, opened 2023-03-23
Fix turtlebot test and improve example

### LINKED ISSUE #397: Re-enable the CI tests for turtlebot once RViz does not randomly crash anymore

ygoumaz:
Fixes #397 together with #686.

This PR fixes the turtlebot3 navigation example. It also restores the corresponding CI test.

The example is now very similar to the TIAGo, with the navigation being started directly from the launch file. The test is also enhanced with a new task checking for the movement of the robot toward a given goal position.

The test is skipped on Rolling because navigation packages are not available. It is also skipped in Foxy, for the same reason as the TIAGo test: the navigation parameters are not the same as on Humble and the distribution is EOF soon.


ygoumaz:
Tests are failing because the Webots nightly package has not been created.

ad-daniel (approved):
Thank you

PR #695, opened 2023-03-30
Fix and cleanup e-puck example

### LINKED ISSUE #694: E-puck example cleanup + fix navigation
Currently, the navigation in the e-puck example doesn't work. Nodes are started but the base-link to map tf is not published. The corresponding test should also be updated, as it should detect such failures.

In addition, the 5 launch files are redundant and can be cleaned-up. The mapper and navigation packages should be started in `robot_launch.py` conditionally. This way, they can be started when the simulation is ready.

- Windows 11
- `webots_ros2` built from sources 68413b6.

ygoumaz:
Fixes #694.

The parameters for the navigation were not up-to-date with Humble. The initial pose is now in the configuration file.

One of the redundant launch files has been deleted. The `epuck-with-tools` test has been improved to better test the navigation.

The rats-life launch files have been updated accordingly.


omichel (approved):

ygoumaz:
The turlebot SLAM test was still crashing on Foxy testing (not main). It is probably coming from an update to one of the package. As Foxy is EOL in May, it is not worth investigating the source of the crash.

omichel (approved):

PR #680, opened 2023-03-22
Fix emitter receiver nodes

angel-ayala:
Changed variables nomenclature for one more standardized and msg new formats.
Update related to #677

PR #696, opened 2023-03-30
Fix Emitter and Receiver devices

angel-ayala:
As continuation PR #680 on adding standard behavior such as #677

### DIFF
--- a/webots_ros2_driver/include/webots_ros2_driver/plugins/static/Ros2Emitter.hpp
+++ b/webots_ros2_driver/include/webots_ros2_driver/plugins/static/Ros2Emitter.hpp
@@ -4,7 +4,7 @@
 #include <webots_ros2_driver/WebotsNode.hpp>
 #include <webots_ros2_driver/plugins/Ros2SensorPlugin.hpp>
 
-#include <webots_ros2_msgs/srv/set_string.hpp>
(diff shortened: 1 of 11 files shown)


ygoumaz (changes requested):
Thank you. I have already tested the behavior and it works nicely. I have two comments on the code and features. Additionally, I don't think `SetInt` is used anywhere in the package. I believe we can remove the `.srv` file too.

ygoumaz (changes requested):
As mentioned in the [Webots documentation](https://cyberbotics.com/doc/reference/receiver#wb_receiver_get_signal_strength), it is illegal to call `wb_receiver_get_signal_strength` and `wb_receiver_get_emitter_direction` if the queue is empty. Therefore, these function calls should be moved inside the loop checking for queue length. A code example can be found here: https://cyberbotics.com/doc/reference/receiver#wb_receiver_get_queue_length.

Moreover, the sensor should be enabled if we subscribe to any of these 3 topics. The `shouldBeEnabled` boolean in `mustPublish` can be changed to take all topics in account and not only `mDataPublisher`.

ygoumaz (approved):
I tested multiple configurations of Emitter and Receiver and all three new topics and it works fine thank you!

PR #705, opened 2023-04-11
Restrict URDF spawner message

ygoumaz:
The `"WARNING: the Ros2Supervisor was not able to spawn this URDF robot"` error message was always triggered, even if the URDF is correctly spawned. This is because the spawner is publishing multiple other messages in stdout than the `success` one.


omichel (approved):
Thank you.

PR #706, opened 2023-04-11
Fix CI tests

ygoumaz:
RViz must be disabled in the CI, to avoid random display failures. The visualization tool is not necessary in the CI.

The `turtlebot3` has no option to disable RViz. The repository is forked and modified to make it optional. The package can be installed from sources in the ROS industrial framework by using the upstream workspace.


omichel (approved):
Thank you.

PR #709, opened 2023-04-14
Add RGB point cloud

lukicdarkoo:
Implements a plugin that merges RGB images from a Camera node and depth images from a RangeFinder node to create a painted point cloud.

image

**Affected Packages**
  - webots_ros2_driver
  - webots_ros2_tests

### DIFF
--- a/webots_ros2/CHANGELOG.rst
+++ b/webots_ros2/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog for package webots_ros2
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+2023.0.4 (2023-XX-XX)
+------------------
+* Added support for painted point clouds
+
 2023.0.3 (2023-04-12)
 ------------------
 * Fixed the calibration of the TIAGo.
--- a/webots_ros2_driver/CHANGELOG.rst
+++ b/webots_ros2_driver/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog for package webots_ros2_driver
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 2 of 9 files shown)


ygoumaz (changes requested):
Thanks for the contribution! I have a few comments in addition to the minor changes suggested:

- The new feature should be added to the main changelog, and also to the sub-package changelogs (driver and tests). You can add the changes under version 2023.0.4.
- I don't understand why 5 bytes are needed for each data point in the PointCloud2 message. There are four pieces of information stored: x, y, z and rgb, all of which are floating point values. Why is rgb stored in byte number 5 instead of number 4? Would it not be possible to reduce each point to 16 bits instead of 20? It seems intentional to do it this way, but I can't figure out the reason.

lukicdarkoo:
> The new feature should be added to the main changelog, and also to the sub-package changelogs (driver and tests). You can add the changes under version 2023.0.4.

Done

> Would it not be possible to reduce each point to 16 bits instead of 20?

Probably not the best reason, but I wanted to match the RealSense implementation.

> For simplicity, we're currently directly unpacking the format that we are 
encountering in the data, which is x,y,z,_,rgb,_,_,_  
each as 32-bit little-endian floats

For Webots, it probably makes sense to pack as `float,float,float,uint8,uint8,uint8` (15 bytes)

ygoumaz:
I believe we can let it this way, as it corresponds to an actual real sensor. Apparently, there is no real convention, even though `float,float,float,uint8,uint8,uint8` is sometimes used.

Thanks for the changelog update. However, the [main changelog](https://github.com/cyberbotics/webots_ros2/blob/feature-rgb-point-cloud/webots_ros2/CHANGELOG.rst) should be updated as well.

ygoumaz (approved):
Very nice, thank you!

ygoumaz (approved):

PR #712, opened 2023-04-18
Cleanup e-puck tools to run RViz alone

ygoumaz:
With the previous e-puck fix (#694), it was not possible to launch RViz from the main launch file without the Navigation or Mapping tool. RViz can still be used alone for odometry visualization for example. This PR cleans up the way tools are launched.


omichel (approved):
Nice.

PR #722, opened 2023-05-02
Calculate vertical fov according to https://www.cyberbotics.com/doc/r…

justinberi:
**Description**
 `Ros2RangeFinder::init` uses the horizontal field of view to determine the vertical focal length. 
According to https://www.cyberbotics.com/doc/reference/rangefinder the vertical FOV (and thus focal length) should be derived from 
`vertical FOV = 2 * atan(tan(fieldOfView * 0.5) * (height / width))`
This PR implements that change.

PS: Thanks for providing this package! It makes working with sim in ROS2 a easy.

**Affected Packages**
List of affected packages:
  - webots_ros2_driver

**Tasks**
  - [x] Implement fix

**Additional context**

Before the fix, point clouds are skewed and do not match the environment.
before

After the fix, point clouds match the environment.
after

### DIFF
--- a/webots_ros2/CHANGELOG.rst
+++ b/webots_ros2/CHANGELOG.rst
@@ -4,6 +4,7 @@ Changelog for package webots_ros2
 
 2023.0.4 (2023-XX-XX)
 ------------------
+* Fixed vertical field of view in ros2 rangefinder plugin.
 * Added support for painted point clouds
 * Fixed ability to launch RViz without other tools in e-puck example.
 * Fixed command line arguments in importer tools.
--- a/webots_ros2/CHANGELOG.rst
+++ b/webots_ros2/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changelog for package webots_ros2
 
 2023.0.4 (2023-XX-XX)
 ------------------
-* Fixed vertical field of view in ros2 rangefinder plugin.
+* Fixed vertical field of view in static RangeFinder plugin.
 * Added support for painted point clouds
 * Fixed ability to launch RViz without other tools in e-puck example.
 * Fixed command line arguments in importer tools.
--- a/webots_ros2_driver/CHANGELOG.rst
+++ b/webots_ros2_driver/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changelog for package webots_ros2_driver
 
 2023.0.4 (2023-XX-XX)
 ------------------
-* Fixed vertical field of view in ros2 rangefinder plugin.
+* Fixed vertical field of view in static RangeFinder plugin.
 * Added support for painted point clouds
 
 2023.0.3 (2023-04-12)
--- a/webots_ros2_driver/CHANGELOG.rst
+++ b/webots_ros2_driver/CHANGELOG.rst
@@ -4,6 +4,7 @@ Changelog for package webots_ros2_driver
 
 2023.0.4 (2023-XX-XX)
 ------------------
+* Fixed vertical field of view in ros2 rangefinder plugin.
 * Added support for painted point clouds
 
 2023.0.3 (2023-04-12)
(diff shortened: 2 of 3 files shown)


webots_ros2_driver/src/plugins/static/Ros2RangeFinder.cpp
```diff
@@ -48,8 +48,10 @@ namespace webots_ros2_driver {
     mCameraInfoMessage.height = height;
     mCameraInfoMessage.width = width;
     mCameraInfoMessage.distortion_model = "plumb_bob";
-    const double focalLengthX = 0.5 * width * (1 / tan(0.5 * wb_range_finder_get_fov(mRangeFinder)));
-    const double focalLengthY = 0.5 * height * (1 / tan(0.5 * wb_range_finder_get_fov(mRangeFinder)));
+    const double hoz_fov = wb_range_finder_get_fov(mRangeFinder);
+    const double vert_fov = 2 * atan(tan(hoz_fov * 0.5) * ( double(height) / width));
+    const double focalLengthX = 0.5 * width * (1 / tan(0.5 * hoz_fov));
+    const double focalLengthY = 0.5 * height * (1 / tan(0.5 * vert_fov));
```
ygoumaz, line 54:
```suggestion
    const double horizontal_fov = wb_range_finder_get_fov(mRangeFinder);
    const double vertical_fov = 2 * atan(tan(horizontal_fov * 0.5) * ( double(height) / width));
    const double focalLengthX = 0.5 * width * (1 / tan(0.5 * horizontal_fov));
    const double focalLengthY = 0.5 * height * (1 / tan(0.5 * vertical_fov));
```

ygoumaz (changes requested):
Thank you for the contribution! Well spotted. I have a few remarks before merging the PR:

- I made some suggestions on the code for clarity.
- You can see that the `test_sources` CI test failed. There is a clang-format error. You can check the logs to see what fails and correct the typo/syntax.
- The changelogs must be updated: [main changelog](https://github.com/justinberi/webots_ros2/blob/fix_rangefinder_point_cloud/webots_ros2/CHANGELOG.rst) and [webots_ros2_driver changelog](https://github.com/justinberi/webots_ros2/blob/fix_rangefinder_point_cloud/webots_ros2_driver/CHANGELOG.rst). You can add an entry for your change under the `2023.0.4` section.

justinberi:
@ygoumaz thanks for your suggested changes! Also updated the change logs (and some more typos :grin:). Let me know if anything else needs fixing.

webots_ros2/CHANGELOG.rst
```diff
@@ -4,6 +4,7 @@ Changelog for package webots_ros2
 
 2023.0.4 (2023-XX-XX)
 ------------------
+* Fixed vertical field of view in ros2 rangefinder plugin.
```
ygoumaz, line 7:
```suggestion
* Fixed vertical field of view in static RangeFinder plugin.
```

webots_ros2_driver/CHANGELOG.rst
```diff
@@ -4,6 +4,7 @@ Changelog for package webots_ros2_driver
 
 2023.0.4 (2023-XX-XX)
 ------------------
+* Fixed vertical field of view in ros2 rangefinder plugin.
```
ygoumaz, line 7:
```suggestion
* Fixed vertical field of view in static RangeFinder plugin.
```

ygoumaz (changes requested):
Just minor suggestions in changelogs and it should be fine!

webots_ros2_driver/CHANGELOG.rst
justinberi: :+1:

ygoumaz (approved):
Thank you!

PR #733, opened 2023-05-08
webots_ros2_driver native MacOS support

HappyEnt:
**Description**
Currently webots_ros2 only builds natively on Ubuntu. Using Robostack, a partially supported ROS2 installation can also be done fairly easily. 
This pull requests makes some changes to the build process, so webots_ros2 also builds natively on MacOS.

**Affected Packages**
List of affected packages:
  - webots_ros2_driver
  - webots_ros2_msgs

**Tasks**
  - [ ] Find better way to build controller library on MacOS
  - [ ] Find better way to resolve dynamic library locations  
  - [ ] fix automatic download of webots

**Additional context**
This pull-request is pretty much WIP and not ready to be merged yet. 
Note that for testing purposes I made changes to the reduced webots source code (the one generated by sync_controller_lib.sh) directly here in the repository.

Currently the build process for MacOS always targets an Application Bundle build, i.e., resources and dynamic libraries are expected to be prefixed with `Contents/...`. As webots_ros2_driver only targets linux for now, it however expects files to be in the unprefixed location. For now I solved this by only conditionally prefixing paths determined by the variable `BUILD_MACOS_BUNDLE`. This however requires changes in the build process of webots. Another solution would be to leave the paths as they are, and change only webots_ros2_driver to use the files under the Contents prefix.

Further note that I bumped the minimum make version to version 3.15 to fix a problem with findPython module during the build process. If this is undesirable I can look If I find a workaround that does not depend on a version bump.

### DIFF
--- a/webots_ros2_driver/CMakeLists.txt
+++ b/webots_ros2_driver/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.15)
 project(webots_ros2_driver)
 
 # Check which ROS distribution is used, ros2control depends of that
--- a/webots_ros2_driver/CMakeLists.txt
+++ b/webots_ros2_driver/CMakeLists.txt
@@ -40,13 +40,17 @@ else()
   find_package(PythonLibs 3.10 EXACT REQUIRED)
 endif()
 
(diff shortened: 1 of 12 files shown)


ygoumaz:
@HappyEnt, do you still have the will to complete this PR?

PR #739, opened 2023-05-09
Fix universal robot example on Humble and Rolling

ygoumaz:
On Humble and Rolling, the universal robot example is not running as smooth as on Foxy and the cans often fall on the ground. This also results in random fails in the corresponding CI test.
See the related issue on the `ros-controls` repository: https://github.com/ros-controls/ros2_controllers/issues/545.

**FOXY**

https://user-images.githubusercontent.com/61198661/237051774-dd9cef4b-85e6-4ed6-a020-bc774b5a8555.mp4

**HUMBLE**

https://user-images.githubusercontent.com/61198661/237051782-0fd4458a-d953-4cba-95b5-170f9ec096fe.mp4

### DIFF
--- a/webots_ros2_control/src/Ros2Control.cpp
+++ b/webots_ros2_control/src/Ros2Control.cpp
@@ -41,7 +41,9 @@ namespace webots_ros2_control {
   void Ros2Control::step() {
     const int nowMs = wb_robot_get_time() * 1000.0;
     const int periodMs = nowMs - mLastControlUpdateMs;
+    // printf("%d - %d = %d\n", nowMs, mLastControlUpdateMs, periodMs);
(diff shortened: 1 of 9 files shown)


ygoumaz:
I am currently trying to fix this issue by adding a velocity control on each joint to limit the speed and get a smoother trajectory. However, this is not the cleanest way to fix it, as it should work only by giving position and the package should be able to interpolate in a smoother way given the times at which positions should be reached.

On another note, the `/state` topic of the trajectory controller is only publishing at a rate of 1Hz. This may be a bug from the `ros-control` package which explains why the interpolation isn't correct.

ygoumaz:
Closing this PR, as the interpolation can be fixed without having to add velocity control to joints. See #768 for follow-up.

PR #760, opened 2023-05-31
Add Iron tests

ygoumaz:
ditto.


ygoumaz:
Iron tests are failing because nav2 + rviz packages are broken.

ygoumaz:
Everything is fixed. In Iron the epuck navigation is only working for orientation but not linear displacement. It may due to a change in the configuration of the nav2 or diffdrive package.

omichel (approved):
Thank you.

Issue #655, opened 2023-03-03
Limitations when resetting simulations


ygoumaz:
In #502, I introduced the possibility to reset all the simulations from the reset button of Webots. However, examples which include tool nodes, like RViz, navigation or others don't work as intended because previous nodes are not shutdown. This includes (at least):

- `webots_ros2_tiago`
- `webots_ros2_turtlebot`
- `webots_ros2_epuck`

ygoumaz:
With the current launch API, there is no way to make the reset work in examples using external nodes like Nav2 and RViz. The reason is that it is mandatory to shutdown all nodes (namely rviz and Nav2 nodes) and restart them when the reset button is pushed. Currently, ROS 2 doesn't allow to shutdown specific nodes from a launch file. The closest possible solution is the following, but it requires to manually restart nodes after pushing the reset button.

1. Webots needs to be started in a specific launch file without other nodes.
2. A second launch file must be started from another process. This launch file contains all other nodes, including robot controllers/plugins, nav2 nodes, rviz, state publishers, etc.
3. The second launch file contains a handler that triggers a shutdown event when the driver node exits (which is the case when the simulation is reset).
4. This second launch file must be manually restarted after pushing the reset button.

PR #764, opened 2023-06-06
Cleanup simulation reset in launch files

ygoumaz:
Follow-up to conclusions in #655.

- `webots_ros2_mavic` and `webots_ros2_tesla` can be simplified to benefit from the `respawn` parameter.
- `webots_ros2_epuck`, `webots_ros2_tiago` and `webots_ros2_turtlebot` are cleaned-up to remove the reset functionality, as there is no clear solution to restart nav2 and rviz nodes. The manual solution with multiple launch files is documented in #655 but won't be implemented in the package. However, it will also be documented in a new tutorial in the official ROS documentation on how to handle the reset button depending on the nodes of the package.
- `webots_ros2_universal_robot` is also dadapted for consistency.

### DIFF
--- a/webots_ros2/CHANGELOG.rst
+++ b/webots_ros2/CHANGELOG.rst
@@ -5,6 +5,7 @@ Changelog for package webots_ros2
 2023.1.0 (2023-XX-XX)
 ------------------
 * Added support for Navigation2 in Iron.
+* Clean simulation reset in launch files.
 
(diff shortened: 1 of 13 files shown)


ygoumaz:
Test failures are not related to this PR, but to a change in Iron and Rolling testing.

omichel (approved):
Thank you.

PR #768, opened 2023-06-07
Fix universal robot interpolation

ygoumaz:
Follow-up to #739.

The interpolation is broken because the /controller-state topic is not published at the correct rate. This is because the configuration files of `ros2_control` are wrong and not working with the current format after Foxy.


webots_ros2_universal_robot/webots_ros2_universal_robot/follow_joint_trajectory_client.py
```diff
@@ -37,8 +38,8 @@ def __init__(self, name, prefix):
         # hotfix for the new topic name in the last version of joint_trajectory_controller
         # (https://github.com/cyberbotics/webots_ros2/pull/726)
         package_xml_path = os.path.join(get_package_share_directory('joint_trajectory_controller'), "package.xml")
-        version = ET.parse(package_xml_path).findall("version")[0].text
-        state_topic = '/state' if version < '3.7.0' else '/controller_state'
+        version_package = version.parse(ET.parse(package_xml_path).findall("version")[0].text)
+        state_topic = '/state' if version_package < version.parse('3.7.0') else '/controller_state'
```
omichel, line 42: That sounds like better English:
```suggestion
        package_version = version.parse(ET.parse(package_xml_path).findall("version")[0].text)
        state_topic = '/state' if package_version < version.parse('3.7.0') else '/controller_state'
```

omichel (changes requested):
Thank you.
I have one minor comment.

omichel (approved):
Thank you.

Issue #476, opened 2022-09-26
SIGINT handling


PhilippPolterauer:
**Bug Description**
when trying to stop a running launch file via CTRL+C the ros driver does not terminate cleanly. 
We encountered this issue when writing a C++ plugin which utilizes RAII and found out that objects which are created inside the C++ plugin are not properly destructed.

**Steps to Reproduce**
1. Install recent version
    1. `mkdir -p ~/dev_ws/src/; cd ~/dev_ws/src; git clone --recursive https://github.com/cyberbotics/webots_ros2.git; cd ..`
    2. `colcon build --symlink-install`
2. Launch a webots simulation (e.g. tesla)
`source install/local_setup.bash; ros2 launch webots_ros2_tesla robot_launch.py`
3. press Ctrl-C
4. observe error below
```
philpolt@vanlinnb12001:~/ros/webots_shutdown_issue$ ros2 launch webots_ros2_tesla robot_launch.py 
[INFO] [launch]: All log files can be found below /home/philpolt/.ros/log/2022-09-26-10-19-37-228324-vanlinnb12001-413857
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [webots-1]: process started with pid [413860]
[INFO] [ros2_supervisor.py-2]: process started with pid [413862]
[INFO] [lane_follower-3]: process started with pid [413864]
[INFO] [driver-4]: process started with pid [413868]
[driver-4] Cannot connect to Webots instance on socket "/tmp/webots-255926/ipc/vehicle/extern", retrying in 0 second...
[ros2_supervisor.py-2] Cannot connect to Webots instance on socket "/tmp/webots-1234/ipc/Ros2Supervisor/extern", retrying in 0 second...
[driver-4] Cannot connect to Webots instance on socket "/tmp/webots-1234/ipc/vehicle/extern", retrying in 1 second...
[ros2_supervisor.py-2] Cannot connect to Webots instance on socket "/tmp/webots-1234/ipc/Ros2Supervisor/extern", retrying in 1 second...
[driver-4] Cannot connect to Webots instance on socket "/tmp/webots-1234/ipc/vehicle/extern", retrying in 2 seconds...
[ros2_supervisor.py-2] Cannot connect to Webots instance on socket "/tmp/webots-1234/ipc/Ros2Supervisor/extern", retrying in 2 seconds...
[...]

PR #781, opened 2023-06-20
wb_robot_cleanup fix for python plugins

angel-ayala:
trying to solve #476 by adding a `stop()` method in python plugins


webots_ros2_driver/src/PythonPlugin.cpp
```diff
@@ -37,6 +42,7 @@ from controller import Supervisor
 sys.path.insert(1, os.path.dirname(controller.__file__))
 from vehicle import Driver
 
+
```
ygoumaz, line 45:
```suggestion
```

ygoumaz (changes requested):
Very nice! Thank you for the contribution. Could you please add an entry in both related changelogs? 

I will further investigate if we can terminate cleanly the supervisor node too.

ygoumaz:
Have you tried to run some examples on Iron distribution (I am not sure if the following error is also on Humble), because I found a problem in examples that don't use Python plugins, namely `webots_ros2_tiago`.

If I start it:
```bash
ros2 launch webots_ros2_tiago robot_launch.py 
```

and then interrupt it with Ctrl+C, I get the following logs:
```logs
[webots_controller_Tiago_Lite-5] Warning: class_loader.ClassLoader: SEVERE WARNING!!! Attempting to unload library while objects created by this loader exist in the heap! You should delete your objects before attempting to unload the library or destroying the ClassLoader. The library will NOT be unloaded.
[webots_controller_Tiago_Lite-5]          at line 127 in ./src/class_loader.cpp
[webots_controller_Tiago_Lite-5] Warning: class_loader.ClassLoader: SEVERE WARNING!!! Attempting to unload library while objects created by this loader exist in the heap! You should delete your objects before attempting to unload the library or destroying the ClassLoader. The library will NOT be unloaded.
[webots_controller_Tiago_Lite-5]          at line 127 in ./src/class_loader.cpp
[webots_controller_Tiago_Lite-5] [ros2run]: Segmentation fault
[ERROR] [webots_controller_Tiago_Lite-5]: process has died [...]
```

It is not related to the changes of this PR but I wonder if we should not try to fix this issue at the same time. What do you think?

angel-ayala:
I have not been able to test it in Iron, I will run it in Humble to see if it also happens.

angel-ayala:
I was checking, that maybe the problem is related to the return added in `WebotsNode.step()`
https://github.com/cyberbotics/webots_ros2/blob/2ce3fb2fca040d72346ded64ff68fc05e13bb284/webots_ros2_driver/src/WebotsNode.cpp#L257
inside the condition when the signal is captured, I'm returning -1 without calling `wb_robot_step` [(code diff)](https://github.com/cyberbotics/webots_ros2/pull/781/files#diff-e4089e8690a190770b4f97a3d72cbfc73cca04b51934f118e633da76436b4971R259)
```python
if (gShutdownSignalReceived && !mWaitingForUrdfRobotToBeRemoved) {
  mRemoveUrdfRobotPublisher->publish(mRemoveUrdfRobotMessage);
  mWaitingForUrdfRobotToBeRemoved = true;
  return -1
}
```
once the SIGINT is received, should it call `wb_robot_step`? this is the only changed behavior that affects both Python and C++ plugins, what do you think @ygoumaz?

ygoumaz:
This error also occurs in the master branch, so it doesn't seem to be related to the changes of this PR. We can already merge this to fix the termination of Python plugins and I will open an issue for C++ plugins on Iron.

ygoumaz:
@angel-ayala, can you update the changelogs, so that we can merge this PR before releasing 2023.1.1?

angel-ayala:
yeah sure, I was finishing some issues with my project's code. I will update this today.

ygoumaz (approved):
Thank you!

PR #789, opened 2023-06-26
Components remapping for robot description

ygoumaz:
Follow-up to https://github.com/cyberbotics/webots/pull/6250.

The driver package allows to get the robot description from the PROTO of the robot and pass it to the state publisher, instead of providing it through a URDF or Xacro file. However PROTO node names may not match the ones of the real robot or the original URDF, making the sim2real conversion complicated. 

This PR introduces a new parameter to provide the path to a YAML file containing a dictionary of names that should remap PROTO names to new values before publishing it to the robot description topic.


omichel (approved):
That looks good to me.
Thank you.

PR #793, opened 2023-06-28
Fix ros2 control hardware interface API update

ygoumaz:
The activation of all components has been deprecated from version 3.15.0 of hardware_interface. It is replaced by manual activation.

This was reported in https://github.com/ros/rosdistro/pull/37344#issuecomment-1609109208.


omichel (approved):
Thank you.

PR #805, opened 2023-07-03
Fix backwards compatibility with driver node

### LINKED ISSUE #796: ROS2 Supervisor unable to connect to Webots instance
## What I tried
1. Install latest release (R2023b) from deb
2. ```
   sudo apt-get install ros-{ROS_DISTRO}-webots-ros2
   ```
3. ```
   ros2 launch webots_ros2_universal_robot multirobot_launch.py
   ```

## What happens
The webots application launches, but the orange robot arm does not move. UR5 model is also not visible. Partial log is as follows:
```
WARNING: No valid Webots directory specified in `ROS2_WEBOTS_HOME` and `WEBOTS_HOME`, fallback to default installation folder /usr/local/webots.
Using load_yaml() directly is deprecated. Use xacro.load_yaml() instead.
Using load_yaml() directly is deprecated. Use xacro.load_yaml() instead.
Using load_yaml() directly is deprecated. Use xacro.load_yaml() instead.
Using load_yaml() directly is deprecated. Use xacro.load_yaml() instead.
Using load_yaml() directly is deprecated. Use xacro.load_yaml() instead.
[INFO] [webots-1]: process started with pid [919714]
[INFO] [ros2_supervisor.py-2]: process started with pid [919716]
[INFO] [ros2-3]: process started with pid [919720]
[INFO] [driver-4]: process started with pid [919722]
[driver-4] Cannot connect to Webots instance, retrying for another 50 seconds...
[ros2_supervisor.py-2] Cannot connect to Webots instance, retrying for another 50 seconds...
```
Eventually the timers run out, and the nodes start over again.

## Additional info
OS: Ubuntu 22.04
ROS2: Humble

In the Webots application console:
```
INFO: 'Ros2Supervisor' extern controller: Waiting for local or remote connection on port 1234 targeting robot named 'Ros2Supervisor'.
```

I also found [this issue](https://stackoverflow.com/questions/69099797/cant-connect-controller-to-specific-instance-of-webots) about setting ```WEBOTS_PID``` but that doesn't seem relevant since I can't predict the PID prior to launch.

ygoumaz:
Fixes #796 (partially).

Since 2023.1.0, it is not possible to declare the driver node in the launch file anymore. Instead the WebotsController should be used instead. This PR restores backwards compatibility and a deprecation message when using the old way.

### DIFF
--- a/webots_ros2_driver/src/Driver.cpp
+++ b/webots_ros2_driver/src/Driver.cpp
@@ -12,8 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <fcntl.h>
 #include <webots/robot.h>
 #include <webots/vehicle/driver.h>
+#include <iostream>
 #include <memory>
 #include <rclcpp/rclcpp.hpp>
 #include <webots_ros2_driver/WebotsNode.hpp>
@@ -38,6 +40,24 @@ int main(int argc, char **argv) {
 
   std::shared_ptr<webots_ros2_driver::WebotsNode> node = std::make_shared<webots_ros2_driver::WebotsNode>(robotName);
   node->init();
+
+  // The parent process must be ros2 run. Declaring from launch file is deprecated. Remove with 2024.0.0.
+  const int BUFFER_SIZE = 4096;
+  const pid_t parentPid = getppid();
+  char processFile[BUFFER_SIZE];
+  sprintf(processFile, "/proc/%d/cmdline", parentPid);
+  int fd = open(processFile, O_RDONLY);
+  char *processName = (char *)malloc(BUFFER_SIZE);
+  int size = read(fd, processName, sizeof(processFile));
+  close(fd);
+  for (int i = 0; i < size; i++) {
+    if (!processName[i])
+      processName[i] = ' ';
+  }
+  if (strstr(processName, "ros2 launch"))
+    RCLCPP_WARN(node->get_logger(), "\033[33mThe direct declaration of the driver node in the launch file is deprecated. "
+                                    "Please use the new WebotsController node instead.\033[0m");
+
   RCLCPP_INFO(node->get_logger(), "Controller successfully connected to robot in Webots simulation.");
   while (true) {
     if (node->step() == -1)
--- a/webots_ros2_driver/CHANGELOG.rst
+++ b/webots_ros2_driver/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog for package webots_ros2_driver
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+2023.1.1 (2023-XX-XX)
+------------------
+* Added deprecation message when declaring driver node in launch file.
+
 2023.1.0 (2023-06-29)
 ------------------
 * Added Ros2Pen static plugin.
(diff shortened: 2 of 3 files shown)


webots_ros2_driver/src/Driver.cpp
```diff
@@ -38,6 +40,24 @@ int main(int argc, char **argv) {
...
   std::shared_ptr<webots_ros2_driver::WebotsNode> node = std::make_shared<webots_ros2_driver::WebotsNode>(robotName);
   node->init();
+
+  // The parent process must be ros2 run. Declaring from launch file is deprecated. Remove with 2024.0.0.
+  const int BUFFER_SIZE = 4096;
+  const pid_t parentPid = getppid();
+  char processFile[BUFFER_SIZE];
+  sprintf(processFile, "/proc/%d/cmdline", parentPid);
+  int fd = open(processFile, O_RDONLY);
+  char *processName = (char *)malloc(BUFFER_SIZE);
```
omichel, line 50: Memory leak: this is never free.
Why do we need to perform a malloc here? Can't we simply reuse the existing buffer (and give it a more generic name)?

webots_ros2_driver/src/Driver.cpp
```diff
@@ -38,6 +40,24 @@ int main(int argc, char **argv) {
 
   std::shared_ptr<webots_ros2_driver::WebotsNode> node = std::make_shared<webots_ros2_driver::WebotsNode>(robotName);
   node->init();
+
+  // The parent process must be ros2 run. Declaring from launch file is deprecated. Remove with 2024.0.0.
+  const int BUFFER_SIZE = 4096;
+  const pid_t parentPid = getppid();
+  char processFile[BUFFER_SIZE];
+  sprintf(processFile, "/proc/%d/cmdline", parentPid);
+  int fd = open(processFile, O_RDONLY);
```
omichel, line 49: We should check the return value and exit with an error message in case the file cannot be open.

omichel (changes requested):

webots_ros2_driver/src/Driver.cpp
```diff
@@ -38,6 +40,27 @@ int main(int argc, char **argv) {
 
   std::shared_ptr<webots_ros2_driver::WebotsNode> node = std::make_shared<webots_ros2_driver::WebotsNode>(robotName);
   node->init();
+
+  // The parent process must be ros2 run. Declaring from launch file is deprecated. Remove with 2024.0.0.
+  const int BUFFER_SIZE = 4096;
+  const pid_t parentPid = getppid();
+  char buffer[BUFFER_SIZE];
```
omichel, line 47:
```suggestion
  const pid_t parentPid = getppid();
  char buffer[4096];
```

omichel (changes requested):

webots_ros2_driver/src/Driver.cpp
```diff
@@ -38,6 +40,27 @@ int main(int argc, char **argv) {
 
   std::shared_ptr<webots_ros2_driver::WebotsNode> node = std::make_shared<webots_ros2_driver::WebotsNode>(robotName);
   node->init();
+
+  // The parent process must be ros2 run. Declaring from launch file is deprecated. Remove with 2024.0.0.
+  const pid_t parentPid = getppid();
+  char buffer[4096];
+  sprintf(buffer, "/proc/%d/cmdline", parentPid);
```
omichel, line 47: `sprintf` is reputed as non-safe and could generate compilation warnings:
```suggestion
  snprintf(buffer, sizeof(buffer), "/proc/%d/cmdline", parentPid);
```

omichel (changes requested):

webots_ros2_driver/src/Driver.cpp
ygoumaz: Yes, I should have seen this one, as I changed all of them in the Webots code recently...

Issue #809, opened 2023-07-05
Supervisor `animation_start_recording` and `animation_stop_recording` services


nikolaseu:
`Ros2Supervisor` is missing the `animation_start_recording` and `animation_stop_recording` services.

These services are documented in the webots website but seem to be available only for ROS 1.

I'm willing to implement them if there's desire to include them officially. I'm currently checking if it actually works (subclassing Ros2Supervisor).

ygoumaz:
This would be a nice contribution. Feel free to open a PR that includes the new services and the required changes in Ros2Supervisor. Don't forget to edit the changelogs with your new updates. We will be happy to review your PR.

PR #810, opened 2023-07-05
Add animation recording services

nikolaseu:
**Description**
Adds the `animation_start_recording` and `animation_stop_recording` services as documented in the [webots documentation](https://cyberbotics.com/doc/reference/supervisor?tab-language=ros#wb_supervisor_animation_start_recording) (for ROS 1).

**Related Issues**
This pull-request fixes issue #809 

**Affected Packages**
List of affected packages:
  - webots_ros2_driver
  - webots_ros2_msgs

**Tasks**
  - [x] Implementation
  - [ ] Documentation ?
  - [ ] Update Changelog

**Additional context**

### DIFF
--- a/webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
+++ b/webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
@@ -37,7 +37,7 @@ from std_msgs.msg import String
 from webots_ros2_driver.utils import is_wsl, has_shared_folder, container_shared_folder, host_shared_folder
 sys.path.insert(1, os.path.join(os.path.dirname(webots_ros2_importer.__file__), 'urdf2webots'))
 from urdf2webots.importer import convertUrdfFile, convertUrdfContent  # noqa
-from webots_ros2_msgs.srv import SpawnUrdfRobot, SpawnNodeFromString  # noqa
+from webots_ros2_msgs.srv import GetBool, SetString, SpawnUrdfRobot, SpawnNodeFromString # noqa
 
 # As Ros2Supervisor needs the controller library, we extend the path here
 # to avoid to load another library named "controller" or "vehicle".
@@ -65,6 +65,9 @@ class Ros2Supervisor(Node):
         # Services
         self.create_service(SpawnUrdfRobot, 'spawn_urdf_robot', self.__spawn_urdf_robot_callback)
         self.create_service(SpawnNodeFromString, 'spawn_node_from_string', self.__spawn_node_from_string_callback)
+        self.create_service(SetString, 'animation_start_recording', self.__animation_start_recording_callback)
+        self.create_service(GetBool, 'animation_stop_recording', self.__animation_stop_recording_callback)
+
         # Subscriptions
         self.create_subscription(String, 'remove_node', self.__remove_imported_node_callback, qos_profile_services_default)
 
@@ -218,6 +221,17 @@ class Ros2Supervisor(Node):
 
         self.get_logger().info('Ros2Supervisor has imported the node named "' + str(object_name) + '".')
         response.success = True
+        return response
+
+    def __animation_start_recording_callback(self, request: SetString.Request, response: SetString.Response):
+        filename = request.value
+        self.get_logger().info(f"Start recording animation to {filename}")
+        response.success = self.__robot.animationStartRecording(filename)
+        return response
+
+    def __animation_stop_recording_callback(self, request: GetBool.Request, response: GetBool.Response):
+        self.get_logger().info(f"Stop recording animation")
+        response.value = self.__robot.animationStopRecording()
         return response
 
     # Allows to remove any imported node (urdf robots / VRML Nodes) by name.
--- a/webots_ros2_msgs/CMakeLists.txt
+++ b/webots_ros2_msgs/CMakeLists.txt
@@ -32,10 +32,11 @@ set(msg_files
   "msg/PenInkProperties.msg"
 )
 set(srv_files
-  "srv/GetBool.srv"
   "srv/EmitterSendString.srv"
-  "srv/SpawnUrdfRobot.srv"
+  "srv/GetBool.srv"
+  "srv/SetString.srv"
   "srv/SpawnNodeFromString.srv"
+  "srv/SpawnUrdfRobot.srv"
 )
 
 rosidl_generate_interfaces(${PROJECT_NAME}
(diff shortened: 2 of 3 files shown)


webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
```diff
@@ -65,6 +65,9 @@ def __init__(self):
         # Services
         self.create_service(SpawnUrdfRobot, 'spawn_urdf_robot', self.__spawn_urdf_robot_callback)
         self.create_service(SpawnNodeFromString, 'spawn_node_from_string', self.__spawn_node_from_string_callback)
+        self.create_service(SetString, 'animation_start_recording', self.__animation_start_recording_callback)
```
nikolaseu, line 68: The services are not namespaced, just like the other services already available in this class. According to the documentation, they should actually be `/supervisor/animation_start_recording`. Should I change it?

webots_ros2_msgs/CMakeLists.txt
```diff
@@ -32,10 +32,11 @@ set(msg_files
   "msg/PenInkProperties.msg"
 )
 set(srv_files
-  "srv/GetBool.srv"
   "srv/EmitterSendString.srv"
-  "srv/SpawnUrdfRobot.srv"
+  "srv/GetBool.srv"
+  "srv/SetString.srv"
   "srv/SpawnNodeFromString.srv"
+  "srv/SpawnUrdfRobot.srv"
```
nikolaseu, line 39: This is just to keep them sorted. It's actually just one line change. Should I revert?

nikolaseu (commented):
@ygoumaz I'm checking the changelog, will push the changes soon, but let me know if this looks ok implementation-wise

webots_ros2_msgs/CMakeLists.txt
ygoumaz: It's nice to have sorted lists, so I believe your change is good

webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
ygoumaz: We should add the namespace to the creation of the `Ros2Supervisor` node. You can add the namespace argument [here](https://github.com/nikolaseu/webots_ros2/blob/enh-record-animation-services/webots_ros2_driver/webots_ros2_driver/webots_launcher.py#L253-L266).

We can use Ros2Supervisor to be as clear as possible which node declared these services.

ygoumaz (commented):
I tried your implementation locally, but unfortunately I could not make it work.
Here is the command I used:

```
ros2 service call /animation_start_recording webots_ros2_msgs/srv/SetString "{value: "test.html" }"
```

The Ros2Supervisor returns the following error and crashes:
```
[ros2_supervisor.py-2] python3: /home/user/ros2_ws/build/webots_ros2_msgs/rosidl_generator_py/webots_ros2_msgs/srv/_set_string_s.c:168: webots_ros2_msgs__srv__set_string__response__convert_from_py: Assertion `PyBool_Check(field)' failed.
```

nikolaseu:
Yes, the return values are not bool, I already had the change locally. Can you try again?

About the namespace, I added it but other things stopped working. Is this what you meant?

```python
class Ros2SupervisorLauncher(Node):
    def __init__(self, output='screen', respawn=True, port='1234', **kwargs):
        # Launch the Ros2Supervisor node
        super().__init__(
            package='webots_ros2_driver',
            executable='ros2_supervisor.py',
            namespace='Ros2Supervisor',
            output=output,
            # Set WEBOTS_HOME to the webots_ros2_driver installation folder
            # to load the correct libController libraries from the Python API
            additional_env={'WEBOTS_CONTROLLER_URL': controller_url_prefix(port) + 'Ros2Supervisor',
                            'WEBOTS_HOME': get_package_prefix('webots_ros2_driver')},
            respawn=respawn,
            **kwargs
        )
```

webots_ros2_driver/webots_ros2_driver/ros2_supervisor.py
ygoumaz: Yes exactly!

ygoumaz:
It works fine on my side when adding the namespace and your last commit fixed the bug of the returned value.

You can add the namespace and update the service name call in the URDF spawner. All tests should pass in the CI.

There is also a source format to fix (see the test source that failed).

You should also update the changelogs of modified packages.

Concerning the documentation, I don't know if you have the rights to modify the wiki? If yes, you can add a new page and add an entry to this one: https://github.com/cyberbotics/webots_ros2/wiki/Tutorial-Ros2Supervisor-Node. If not, I will take care of it.

nikolaseu:
Perfect. I pushed a fix for the failing test, added the namespace, and for the changelogs I followed the same style as #789

nikolaseu:
For me it's still not working, nav2 cannot find the transformation between robot and map. The only thing I see in the logs is (for every robot):

```
[ros2_supervisor.py-2] Warning: Webots [R2023b] and libController [R2023b revision 1] versions are not the same for Robot 'Ros2Supervisor'! Different versions can lead to undefined behavior.
```

I'm using Ubuntu 22.04 with latest release R2023b

ygoumaz:
I also can reproduce the problem. Let's see if this is also the case in the CI. I will try to find the source of it. It may be linked to topics published by the supervisor that are affected by the new namespace.

ygoumaz:
The problem can be fixed by remapping the clock topic in the Ros2Supervisor. You can add the following option to the creation of the Ros2Supervisor node.
```python
remappings=[('/Ros2Supervisor/clock', '/clock')],
```

ygoumaz (approved):
Thank you, it works fine and tests pass!

Concerning the wiki documentation, are you allowed to modify the wiki pages?

nikolaseu:
Yes, I tried and it was saved. It's a bit unfortunate that there's no review process for the wiki though. Here you can see my changes: https://github.com/cyberbotics/webots_ros2/wiki/Tutorial-Ros2Supervisor-Node/_compare/b4286a40769b1790483ba8f302080edb8a341869...e3ec0fce004526138458b958039486e11489e000

nikolaseu:
Thanks Yannick! I can't merge though, I don't have write access to the repo. So feel free to press the button yourself.

ygoumaz:
You have been added to the committers of the repo! Once the tests pass, you will be able to merge it yourself.

PR #819, opened 2023-07-07
Enhancement Add PROTOSpawner

### LINKED ISSUE #818: New PROTOSpawner Feature
From version 2024.0.0, importing URDF and Xacro files will be deprecated (see #815). It is now already possible to import robot strings using the Ros2Supervisor (which uses the `importMFNodeFromString` API function).

However, this feature is very limited as PROTOs can't be imported by providing a file path and can't be declared in the robot string without declaring their URLs as IMPORTABLE in the world file.

### Implementation blueprint
The new `PROTOSpawner` node should replace the current `URDFSpawner`. It should take either a PROTO path or a robot string as argument, as well as a list of URLs to add to the IMPORTABLE list. The current robot_string service created by the `Ros2Supervisor` should be enhanced to be called with those arguments.

The `Ros2Supervisor` (in the service callback) is then responsible to add the URLs to the IMPORTABLE list (using a new API function implemented in #[...]). In the same way as the `URDFSpawner` callback, it must update intern URLs and copy PROTO files (if needed) to the correct location in function of the platform (WSL, macOS, Linux).

ygoumaz:
Resolves #818.

TODO: describe full implementation.

### DIFF
--- a/webots_ros2_msgs/srv/SpawnProtoRobot.srv
+++ b/webots_ros2_msgs/srv/SpawnProtoRobot.srv
@@ -1,0 +1,3 @@
+UrdfRobot robot
+---
+bool success
--- a/webots_ros2_driver/webots_ros2_driver/proto_spawner.py
+++ b/webots_ros2_driver/webots_ros2_driver/proto_spawner.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python
+
+# Copyright 1996-2023 Cyberbotics Ltd.
+#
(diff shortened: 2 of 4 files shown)


webots_ros2_msgs/srv/SpawnProtoRobot.srv
```diff
@@ -0,0 +1,3 @@
+UrdfRobot robot
```
IDavGal, line 1:
```suggestion
ProtoRobot robot
```

PR #824, opened 2023-07-11
Fix Ros2Supervisor import and remove node strings

ygoumaz:
While writing the `Ros2Supervisor` [tutorial](https://github.com/cyberbotics/ros2_documentation/pull/18), I came across some bugs in the import and remove functions of the supervisor node.

The name of the imported nodes should be able to contain capital letters and the remove function was not deleting the correct node in some scenarios.


omichel (approved):
Thank you.

PR #841, opened 2023-08-08
add remappings to robot state publisher for ur5e example

skpawar1305:
follow up to https://github.com/cyberbotics/webots_ros2/commit/f3045ad16c6c41b0edbf845f06391be5d0990022

**Description**
robot state publisher's robot_description update service doesn't get called, and also doesn't subscribe to joint_states' topic, because of the namespace 'ur5e' added to the WebotsController
add remappings to the robot state publisher node to fix it

**Affected Packages**
List of affected packages:
  - webots_ros2_universal_robot

### DIFF
--- a/webots_ros2_universal_robot/launch/robot_launch/robot_moveit_nodes_launch.py
+++ b/webots_ros2_universal_robot/launch/robot_launch/robot_moveit_nodes_launch.py
@@ -41,6 +41,13 @@ def generate_launch_description():
 
     # Check if moveit is installed
     if 'moveit' in get_packages_with_prefixes():
+        # Webots simulation with robot
+        launch_description_nodes.append(
+            IncludeLaunchDescription(
+                PythonLaunchDescriptionSource(os.path.join(package_dir, 'launch', 'robot_nodes_launch.py'))
+            )
+        )
+
         # Configuration
         description = {'robot_description': load_file('ur5e_with_gripper.urdf')}
         description_semantic = {'robot_description_semantic': load_file('moveit_ur5e.srdf')}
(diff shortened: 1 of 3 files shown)


skpawar1305:
adding a namespace to the node should've been better, but webotscontroller's service client calls incorrect topic 
'/ur5erobot_state_publisher/set_parameters' instead of '/ur5e/robot_state_publisher/set_parameters'

omichel:
In that case, I believe we should fix the webots controller service client to call the correct topic.

webots_ros2_universal_robot/launch/robot_launch/robot_moveit_nodes_launch.py
```diff
@@ -85,14 +92,7 @@ def load_yaml(filename):
...
                 ],
-                remappings=[('/joint_states', '/ur5e/joint_states')]
-            )
-        )
-
-        # Webots simulation with robot
-        launch_description_nodes.append(
-            IncludeLaunchDescription(
-                PythonLaunchDescriptionSource(os.path.join(package_dir, 'launch', 'robot_nodes_launch.py'))
+                remappings=[('/joint_states', '/ur5e/joint_states')],
```
omichel, line 95:
```suggestion
                remappings=[('/joint_states', '/ur5e/joint_states')]
```

omichel (changes requested):
Do you know why the tests are failing on rolling?

skpawar1305:
> Do you know why the tests are failing on rolling?

no

omichel (approved):
I manually restarted the rolling test and it passed.
Probably a random failure or something that was fixed meanwhile.

PR #846, opened 2023-09-02
implemented perform_command_mode_switch override in Ros2ControlSystem

ralphieraccoon:
**Description**
Reusing the solution from [here](https://github.com/ros-controls/gazebo_ros2_control/pull/136/files) for Gazebo to allow for multiple command interfaces. Implements the `perform_command_mode_switch` method override to set joint control methods when controllers are activated/deactivated. Can't work out how to detect if a command interface is claimed inside `Ros2ControlSystem::init()` so you either have to run `switch_controllers` once to call the function or start with a position controller.

**Related Issues**
This pull-request references the discussion #845 I started earlier.

**Affected Packages**
List of affected packages:
  - webots_ros2_control

**Tasks**
  - [ ] Work out how to detect if a command interface is claimed on `Ros2ControlSystem::init()` so it's all set up properly at the beginning without having to run `switch_controller`.

**Additional context**
Credit to @ksotebeer for the original PR for `gazebo_ros2_control`.

### DIFF
--- a/webots_ros2_control/include/webots_ros2_control/Ros2ControlSystem.hpp
+++ b/webots_ros2_control/include/webots_ros2_control/Ros2ControlSystem.hpp
@@ -61,6 +61,9 @@ namespace webots_ros2_control {
 
     std::vector<hardware_interface::StateInterface> export_state_interfaces() override;
     std::vector<hardware_interface::CommandInterface> export_command_interfaces() override;
+
+    hardware_interface::return_type perform_command_mode_switch(const std::vector<std::string>& start_interfaces, const std::vector<std::string>& stop_interfaces) override;
+
     hardware_interface::return_type read(const rclcpp::Time & /*time*/, const rclcpp::Duration & /*period*/) override;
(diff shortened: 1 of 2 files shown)


ralphieraccoon:
I'm not sure if discovering which command interface is claimed inside `Ros2ControlSystem::init()` is possible since it doesn't provide access to the `ResourceManager` which has this function. To make it work might require changes to the `ros2_control` base class. I can take this PR out of draft as is and the problem could be solved at a later date, what would be best?

omichel:
I would prefer that you test and eventually solve the problem in this PR, rather than committing something which is unfinished or buggy.

ralphieraccoon:
> I would prefer that you test and eventually solve the problem in this PR, rather than committing something which is unfinished or buggy.

That's fair, but at the moment I'm not sure the issue is fixable without alterations to `ros2_control`. The PR doesn't break existing functionality, if using a single command interface is used it works the same as before, and the issue can be worked around by calling `switch_controllers`. If anyone has any suggestions on how to solve the issue, it would be greatly appreciated.

Issue #759, opened 2023-05-24
No external controller connection(ROS2 on WLS to Webots on Windows)


Dan010123:
**Describe the Bug**
launched ROS2 Webot demos open Webots, show the world but stuck on waiting for extern controller connection on IPC or TCP which newer happens.

**Steps to Reproduce**
1. Installed ROS2  + ros-humble-webots-ros2 on WSL Ubuntu 22.04
2. Installed Webots on Windows10
3. Configured workspace, sourced ROS, exported WEBOTS_HOME=/mnt/d/Programs/Webots
4. Started  (for example) ros2 launch webots_ros2_turtlebot robot_launch.py

**Expected behavior**
Simulation running

**Screenshots**
![изображение](https://github.com/cyberbotics/webots/assets/134538828/4afdaa41-2f29-467e-be4f-a0f78b573d36)
![изображение](https://github.com/cyberbotics/webots/assets/134538828/1477d051-de31-4f9d-b826-f1b6c9902012)

**System**
 - WIndows 10 + Ubuntu 22.04 on WSL

**Additional context**
I guess I should somehow get IPC or TCP configured, but can't find info on how to manage this....
Thanks!

ygoumaz:
I will take a look at your issue. It is possible that the package is somehow broken on Windows.

Could you just give me a few more details:

- Have you built the ROS 2 package (webots_ros2) from sources or installed from apt binaries?
- What version of the package are you using? You can execute the following command to get it: `ros2 pkg xml webots_ros2 | grep version`.

suresh-venkate:
> I will take a look at your issue. It is possible that the package is somehow broken on Windows.
> 
> Could you just give me a few more details:
> 
> * Have you built the ROS 2 package (webots_ros2) from sources or installed from apt binaries?
> * What version of the package are you using? You can execute the following command to get it: `ros2 pkg xml webots_ros2 | grep version`.

In my case, I have installed from apt binaries using this command: 'sudo apt-get install ros-$ROS_DISTRO-webots-ros2'. (ROS_DISTRO is humble). 

The package version that I see is ' <version>2023.0.3</version>'

Thanks

ygoumaz:
[...]

PR #879, opened 2024-01-10
support Mirrored mode networking for wsl2

hellototoro:
**Description**
[Mirrored mode networking](https://learn.microsoft.com/en-us/windows/wsl/networking#mirrored-mode-networking) is a new network mode of wsl2. In this mode, we can use localhost accessing Windows networking apps from Linux.

When the following is written in the `.wslconfig` file, the wsl2 is switched to Mirrored mode
```
# Enable experimental features
[experimental]
# Network
networkingMode=mirrored
dnsTunneling=true
firewall=true
autoProxy=true
```
**Related Issues**
maybe [#759](https://github.com/cyberbotics/webots_ros2/issues/759#issuecomment-1878485262)

**Affected Packages**
  - webots_ros2_driver

**Tasks**
  - [ ] support Mirrored mode networking for wsl2
  - [ ] rename a few function

**Additional context**

### DIFF
--- a/webots_ros2_driver/webots_ros2_driver/utils.py
+++ b/webots_ros2_driver/webots_ros2_driver/utils.py
@@ -139,7 +139,8 @@ def get_wsl_ip_address():
     # wsl2 net mode
     # NAT: if get_router_ip() == get_wsl_host_ip()
     # Mirrored: if get_router_ip() != get_wsl_host_ip()
-    return get_wsl_host_ip() if get_router_ip() == get_wsl_host_ip() else '127.0.0.1'
+    wsl_host_ip = get_wsl_host_ip()
+    return wsl_host_ip if get_router_ip() == wsl_host_ip else '127.0.0.1'
 
 
 def controller_protocol():
--- a/webots_ros2_driver/webots_ros2_driver/utils.py
+++ b/webots_ros2_driver/webots_ros2_driver/utils.py
@@ -86,7 +86,7 @@ def is_wsl():
     return 'microsoft-standard' in uname().release
 
 
-def get_wsl_ip_address():
+def get_wsl_host_ip():
     try:
         file = open('/etc/resolv.conf', 'r')
     except IOError:
@@ -123,7 +123,7 @@ def container_shared_folder():
     return shared_folder_list[1]
 
 
-def get_host_ip():
+def get_router_ip():
     try:
         output = subprocess.run(['ip', 'route'], check=True, stdout=subprocess.PIPE, universal_newlines=True)
         for line in output.stdout.split('\n'):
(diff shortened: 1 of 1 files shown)


webots_ros2_driver/webots_ros2_driver/utils.py
```diff
@@ -135,19 +135,26 @@ def get_host_ip():
...
+
+
 def controller_protocol():
     protocol = 'tcp' if (has_shared_folder() or is_wsl()) else 'ipc'
     return protocol
 
 
 def controller_ip_address():
-    ip_address = get_host_ip() if has_shared_folder() else get_wsl_ip_address()
+    ip_address = get_router_ip() if has_shared_folder() else get_wsl_ip_address()
```
omichel, line 151:
```suggestion
    ip_address = get_router_ip() if has_shared_folder() else get_wsl_host_ip()
```

webots_ros2_driver/webots_ros2_driver/utils.py
```diff
@@ -135,19 +135,26 @@ def get_host_ip():
...
     return ip_address
 
 
 def controller_url_prefix(port='1234'):
     if has_shared_folder() or is_wsl():
-        return 'tcp://' + (get_host_ip() if has_shared_folder() else get_wsl_ip_address()) + ':' + port + '/'
+        return 'tcp://' + (get_router_ip() if has_shared_folder() else get_wsl_ip_address()) + ':' + port + '/'
```
omichel, line 157:
```suggestion
        return 'tcp://' + (get_router_ip() if has_shared_folder() else get_wsl_host_ip()) + ':' + port + '/'
```

webots_ros2_driver/webots_ros2_driver/utils.py
```diff
@@ -135,19 +135,26 @@ def get_host_ip():
         sys.exit('Unable to get host IP address. \'ip route\' could not be executed.')
 
 
+def get_wsl_ip_address():
+    # wsl2 net mode
+    # NAT: if get_router_ip() == get_wsl_host_ip()
+    # Mirrored: if get_router_ip() != get_wsl_host_ip()
+    return get_wsl_host_ip() if get_router_ip() == get_wsl_host_ip() else '127.0.0.1'
```
omichel, line 142: We should cache the result of calling this method as it is not lightweight:
```suggestion
    wsl_host_ip = get_wsl_host_ip()
    return wsl_host_ip if get_router_ip() == wsl_host_ip else '127.0.0.1'
```

omichel (changes requested):
Thank you for this contribution.
However, I would request a couple of changes.

webots_ros2_driver/webots_ros2_driver/utils.py
hellototoro: you're right

webots_ros2_driver/webots_ros2_driver/utils.py
hellototoro: but get_wsl_ip_address shoud be used here

hellototoro: if the file `/etc/resolv.conf` exist, the function get_wsl_host_ip never return `127.0.0.1` whatever the mode.
But get_wsl_ip_address can return get_wsl_host_ip or localhost

omichel: What I mean here is that in this PR, `get_wsl_ip_address` doesn't exist any more. You renamed it to `get_wsl_host_ip`. So, the above line will fail due to an attempt to call an unknown function.

PR #896, opened 2024-02-14
set is_urdf_loaded__ of the resource manager to true

skpawar1305:
it's probably needed after https://github.com/ros-controls/ros2_control/commit/385293e5d0626c57636a05c66de4380fcfd5bee7

fixes https://github.com/cyberbotics/webots_ros2/issues/893

**Description**
After, https://github.com/ros-controls/ros2_control/commit/385293e5d0626c57636a05c66de4380fcfd5bee7, init_services() won't be called unless is_urdf_loaded__ is set to true. Passing robot description again led to controller name duplicate error.

**Related Issues**
This pull-request fixes issue https://github.com/cyberbotics/webots_ros2/issues/893

**Affected Packages**
List of affected packages:
  - webots_ros2_control

**Tasks**
Add the list of tasks of this PR.
  - [x] Fix ros2_control usage for ros2 humble

PR #898, opened 2024-02-15
Fixed #896 so it works for ROS2 Iron.

ralphieraccoon:
**Description**
Quick bugfix for #896, which doesn't work for `iron` or `rolling` because they use a `ros2_control` hardware interface version  `>3.15`. For some reason still fails for `rolling`. 

**Affected Packages**
List of affected packages:
  - webots_ros2_control


omichel (approved):
Nice, thank you!

PR #939, opened 2024-07-05
Webots now handles initial_value params in position state_interface.

ralphieraccoon:
**Description**
If the `initial_value` param in the position `state_interface` URDF is set, then the robot joint will move to that position when Webots starts.

**Affected Packages**
List of affected packages:
  - webots_ros2_control

**Additional context**
Currently uses `wb_motor_set_position` which means the robot will start at zero and instantly move once Webots finishes loading. Tried to implement the same behavior using `wb_supervisor_node_set_joint_position` but with no success.


omichel (approved):
Thank you.

PR #1025, opened 2025-01-10
Update Resource Manager API

destogl:
Update of ResouceManager API from ros2_control that is proposed in ros-controls/ros2_control#1049


lukicdarkoo:
@destogl Thank you for the contribution and https://github.com/ros-controls/ros2_control/pull/1049 is a feature we really needed on some projects.

Is this change going to hit Jazzy or Rolling only?

PR #1026, opened 2025-01-16
Adding Connector Plugin

ralphieraccoon:
**Description**
Integrating the Webots `Connector` node into `webots_ros2`.

**Affected Packages**
  - webots_ros2_driver
  - webots_ros2_msgs

### DIFF
--- a/webots_ros2_driver/CMakeLists.txt
+++ b/webots_ros2_driver/CMakeLists.txt
@@ -32,6 +32,8 @@ find_package(yaml-cpp REQUIRED)
 
 if($ENV{ROS_DISTRO} MATCHES "humble")
   find_package(Python 3.10 EXACT REQUIRED COMPONENTS Development)
+elseif($ENV{ROS_DISTRO} MATCHES "iron")
+  find_package(Python 3.10 EXACT REQUIRED COMPONENTS Development)
 elseif($ENV{ROS_DISTRO} MATCHES "jazzy")
   find_package(Python 3.12 EXACT REQUIRED COMPONENTS Development)
 elseif($ENV{ROS_DISTRO} MATCHES "rolling")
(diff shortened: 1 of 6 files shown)


skpawar1305:
I don't know anything about Webots Connector, please choose someone else as reviewer @ralphieraccoon

ralphieraccoon:
"No clock received, and time argument is zero" seems to be the error causing the tests to fail. I'm getting it on my local build now too. Has a package update broken `webots_ros2`?

lukicdarkoo (changes requested):
@ralphieraccoon Thank you, this looks quite good.

For wb_connector_lock() I would rather use a service, but this is fine too. We can add a service as well later.

Could you please add a test for this?

ralphieraccoon:
No problem! I can write a service if you prefer, I used `VacuumGripper` as a template. I guess you might want a service for that too?

As for tests, is there a particular robot you'd like to use? I guess I could put a tool changer on a robot arm.

lukicdarkoo:
@ralphieraccoon 

> No problem! I can write a service if you prefer, I used VacuumGripper as a template. I guess you might want a service for that too?

This is a good point, let's keep consistent with the vacuum gripper.

> As for tests, is there a particular robot you'd like to use? I guess I could put a tool changer on a robot arm. Speaking of which, I think VacuumGripper is missing a test too.

Whatever robot you prefer. Please add a test for connectors only in this PR. If you want, you can open a separate PR for the vacuum gripper.

ralphieraccoon:
OK, what I think I'll do is add a world to the `webots_ros2_universal_robot` that includes a tool changer that can used to test the `Connector` node (which I'll write a test for in `test_system_universal_robot.py`). Then in another PR I'll add a vacuum gripper and add another test in there for that.

lukicdarkoo:
If adding a new world please add it in the webots_ros2_tests package and keep it as simple as possible. It can be UR robot or it can be the one we use for driver tests.

PR #1027, opened 2025-01-27
Guard condition to check node clock is not zero, to suppress "No clock received" error from ros2_control.

ralphieraccoon:
**Description**
This adds a guard condition to `Ros2Control::step()` to check node clock is not zero, fixes the "No clock received" error. Basically stops a step update occurring unless the clock is not zero.

**Affected Packages**
List of affected packages:
  - webots_ros2_control

### DIFF
--- a/webots_ros2_control/src/Ros2Control.cpp
+++ b/webots_ros2_control/src/Ros2Control.cpp
@@ -90,7 +90,7 @@ namespace webots_ros2_control {
   void Ros2Control::step() {
     const int nowMs = wb_robot_get_time() * 1000.0;
     const int periodMs = nowMs - mLastControlUpdateMs;
-    if (periodMs >= mControlPeriodMs) {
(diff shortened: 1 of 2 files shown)


skpawar1305:
this might cause issues when simulation is running at realtime, can you check?

ralphieraccoon:
> this might cause issues when simulation is running at realtime, can you check?

I have a physical robot arm I could test it on, give me a day or two 🙂

ralphieraccoon:
I ran `ros2 launch webots_ros2_epuck robot_launch.py "use_sim_time:=false"` and sent a command to `/cmd_vel`. The video is below, it looks the same as if I run it with `"use_sim_time:=true"` 

https://github.com/user-attachments/assets/ea6e8b13-256a-4be6-a545-75de7c6a8514

Does that look right to you?

skpawar1305 (approved):

Issue #1031, opened 2025-02-16
Webots Ros2 program (any of the demos) running in Ubuntu (Windows WSL2) is not able to connect to Webots host simulator in Windows 11.


andrespineda:
**Describe the Bug**
After installing Webots on a Windows 11 host configured with a Ubuntu 22.04 on WSL2, any of the demo programs fail in trying to connect to the Windows Webot simulator. 

**Steps to Reproduce**
1. Install the Ros2 and Webots code as specified in the Ros2 documentation: https://docs.ros.org/en/foxy/Tutorials/Advanced/Simulators/Webots/Installation-Windows.html#install-wsl2https://docs.ros.org/en/foxy/Tutorials/Advanced/Simulators/Webots/Installation-Windows.html#install-wsl2.
2. After installation, attempt to run the provided demo script:  ros2 launch webots_ros2_universal_robot multirobot_launch.py.

**Expected behavior**
The expectation is that the code will start to run and pop open the Webots simulator window on the Windows host. It will them start to send commands to the robot running in the simulation to perform the demo actions.
Instead, the program starts up, but is unable to connect to the host. The screen continuously displays an error that it cannot connect and will try in a few seconds. It never is able to connect.
 
**Affected Packages**
List of affected packages:
 All of the demo programs, such as epuck, universal_robot, etc.

**Screenshots**
Not available

**System**
 - Webots Version: [R2025a ]
 - ROS Version: [Jazzy]
 - Operating System: [Windows 11, WSL2 Ubuntu 24.04.2 LTS]
 - Graphics Card: [Unknown]

**Additional context**

I spent over a full day trying to figure this out and found a solution.

In my case, the issue was that the program located in "webots_ros2_driver/lib/python3.12/site-packages/webots-driver/utils.py" has a function named "get_wsl_ip_address()" . This function gets the "nameserver" ip address from the WSL2  "/etc/resolv.conf" file and returns it incorrectly as the ip address of the Windows host. Because of this, the Ubuntu webots controller keeps trying to connect to the webots simulator running on the Windows side using the nameserver ip address as the TCP IP address. I my system, the nameserver was set to 127.0.0.53. The "webots-controller"  kept trying to connect to TCP 127.0.0.53 port 1234.

To prove that this was the problem, I manually edited the "/etc/resolv.conf" file and changed the nameserver address to "127.0.0.1".

[...]

PR #1032, opened 2025-03-12
Fix WSL IP detection when DNS Tunnelling is enabled

Rami-Sabbagh:
**Description**
On a fresh WSL Ubuntu 24.04 installation, ROS2 Jazzy and latests `webots_ros2`. The package fails to connect with the launched Webots instanced.

Upon digging up it turns out the issue is using the wrong IP to connect with the Windows host.

The implementation looks at the ip in `/etc/resolv.conf`, which used to be the Windows host IP, but that's no longer the case when the DNS Tunneling feature is enabled (which is the case by default for new WSL installations).

When DNS Tunneling is enabled, the IP in `/etc/resolv.conf` becomes `10.255.255.254`, which is configured as a lookback address.

The solution I have implemented is to fallback to `get_host_ip` when this IP is found. `get_host_ip` uses `ip route` to get the gateway IP and use it, which is the right one as long as WSL NAT networking mode is used.

This leaves only the case when bot Mirrored networking & NAT Tunneling are enabled together. In such case the package would fail to connect, it should use the IP `127.0.0.1` for it (loopback address).

**Related Issues**
This pull-request fixes issue #1031.

**Affected Packages**
  - webots_ros2_driver

**Tasks**
  - [x] Fix WSL IP detection when DNS Tunnelling is enabled.

**Additional context**
Checkout my [detailed comment] in #1031.

### DIFF
--- a/webots_ros2_driver/webots_ros2_driver/utils.py
+++ b/webots_ros2_driver/webots_ros2_driver/utils.py
@@ -102,6 +102,8 @@ def get_wsl_ip_address():
                 file.close()
                 if len(tokens[1]) == 0:
                     return '127.0.0.1'
+                if tokens[1] == '10.255.255.254':
+                    return get_host_ip()
                 return tokens[1]
     finally:
         file.close()


omichel (approved):
Thank you!

PR #1035, opened 2025-03-20
Use Image Transport

skpawar1305:
- **use image_transport library to publish color images**
- **use image_transport library to publish depth images**

**Description**
Use image_transport library to publish images

**Affected Packages**
List of affected packages:
  - webots_ros2_driver

### DIFF
--- a/webots_ros2_driver/CMakeLists.txt
+++ b/webots_ros2_driver/CMakeLists.txt
@@ -19,6 +19,7 @@ find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_python REQUIRED)
 find_package(rosgraph_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
+find_package(image_transport REQUIRED)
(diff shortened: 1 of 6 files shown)


lukicdarkoo:
@skpawar1305 What is the reason for using image_transport over regular raw images? Do you have a specific use case?

skpawar1305:
> @skpawar1305 What is the reason for using image_transport over regular raw images? Do you have a specific use case?

In our student course, one of the tasks is to command a robot in another person's computer using the image feed

as you know the raw image size is too much for a network to handle, they can simply use compressed image

last year we simply skipped this task, this time maybe we do it..

skpawar1305:
@lukicdarkoo hello, please close this PR if it seems unnecessary 
thank you

lukicdarkoo:
@skpawar1305 I see. I believe a better solution is to add the `image_transport/republish` node in your launch file and let it handle the compression.

As there is an alternative solution I will close the PR.

PR #1039, opened 2025-06-11
Replace ament_target_dependencies with target_link_libraries

ahcorde:
`ament_target_dependencies` is deprecated on `kilted` and removed in `rolling`,  It will require a new release on `rolling`.


lukicdarkoo:
@ahcorde Thank you for the PR. I believe we still need to use ament_target_dependencies for Humble and Jazzy?

PR #1053, opened 2026-05-25
Set GPS NavSatFix status

jkaflik:
## Motivation
 already publishes latitude, longitude, altitude, and , but leaves  at its default  value. Downstream localization stacks commonly reject  messages with negative status values, treating them as invalid/no-fix data even though Webots is publishing a valid simulated GPS fix.


omichel (approved):
The test are failing, but this is not due to this change.

